PR #1, opened 2017-03-07
Improve the syntax, markup and fix minor glitches in the README

MattiSG:


MattiSG:
Ah ok, never mind, I understood how you recommend using this. 8f34210 should probably be removed from this PR. I don't really see the point of `init + pull` vs `clone`, though.

I guess it's because you don't want `origin` to be set`. If that's the case, I would find the intention clearer if you did a `clone` and a `remote remove`.

Another option, which would sound as legit (if not more) to me, would be to download a dump and `git init` from it.

PR #7, opened 2017-05-24
Add reform modifying brackets of a scale

cbenz:
Connected to https://github.com/openfisca/openfisca-france/issues/745

* Tax and benefit system evolution: add a reform modifying the brackets of a scale
* Impacted periods: from 2013-01-01
* Impacted areas: `openfisca_country_template/parameters/taxes.xml`
* Details:
  - Show how to add, modify and remove a bracket.
  - Add corresponding tests.

- - - -

These changes:

- Add a new feature: a reform

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,16 @@
+# 1.1.0 - [#7](https://github.com/openfisca/country-template/pull/7)
+
+* Tax and benefit system evolution: add a reform modifying the brackets of a scale
+* Impacted periods: from 2013-01-01
+* Impacted areas: `openfisca_country_template/parameters/taxes.xml`
+* Details:
+  - Show how to add, modify and remove a bracket.
+  - Add corresponding tests.
+
 # 1.0.0 - [#4](https://github.com/openfisca/country-template/pull/4)
 
 * Tax and benefit system evolution.
-* Impacted periods: all. 
+* Impacted periods: all.
 * Impacted areas:
   - `benefits`
   - `demographics`
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ These elements are described in different folders. All the modelling happens wit
 
 - The rates are in the `parameters` folder.
 - The formulas are in the `variables` folder.
+- This country template comes also with *reforms* in the `reforms` folder. This is optional: your country may exist without defining any reform.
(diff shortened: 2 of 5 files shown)


openfisca_country_template/reforms/modify_social_security_brackets.py
```diff
@@ -0,0 +1,43 @@
...
+# This file defines a reform.
+# A reform is a set of modifications to be applied to a reference tax and benefit system to carry out experiments.
+# See https://doc.openfisca.fr/reforms.html
+
+
+# Import from openfisca-core the common python objects used to code the legislation in OpenFisca
+from openfisca_core.model_api import *
+
+
+class modify_social_security_brackets(Reform):
```
MattiSG: This describes what is done, and that's not usually how reforms are named: they usually are named after the law project, or the impact they want to have. These examples are meaningful only if one can rely to it on business cases.

So, here, what would have happened in the modelled country? Refer to the README and add to the story there: maybe this reform was a law proposal made by some representatives as an alternative to the universal basic income?

openfisca_country_template/tests/reforms/modify_social_security_brackets.yaml
```diff
@@ -0,0 +1,31 @@
+# Test files describe situations and their expected outcomes
+# We can run this test on our command line using `openfisca-run-test social_security_contribution.yaml`
+
+# This test is a partial adaptation of `social_security_contribution.yaml` to match the modifications introduced
+# by the reform of the same name.
```
MattiSG: I think we rely on line wrapping on other comments. If that's the case, please be consistent.

openfisca_country_template/tests/reforms/modify_social_security_brackets.yaml
```diff
@@ -0,0 +1,30 @@
+# Test files describe situations and their expected outcomes
+# We can run this test on our command line using `openfisca-run-test social_security_contribution.yaml`
+
+# This test is a partial adaptation of `social_security_contribution.yaml` to match the modifications introduced by the reform of the same name.
+
+# Note the `reforms: ` key in the below YAML blocks.
+
+- name: Person earning 2000 in 2013-01
```
MattiSG: The date doesn't match :wink:

But I guess that shows the name is bad if it is just a copy of data that is readily available below. What about something like “No social security contribution on small salaries”?

(repeat below)

PR #15, opened 2017-07-31
Check version number change before deployment 

Anna-Livia:
Connected to openfisca/openfisca-core#513

This PR is the implementation in the Country Template Package of the procedure to check version number before attempting to deploy discussed in  [Check version number change before deployment](https://github.com/openfisca/openfisca-core/pull/547).

#### Technical changes
When there is a new commit on the master branch, if the version number has not been changed, the deployment will not be triggered.

### DIFF
--- a/deploy-if-version-bump.sh
+++ b/deploy-if-version-bump.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+set -e
+if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+    git tag `python setup.py --version`
+    git push --tags  # update the repository version
+    python setup.py bdist_wheel  # build this package in the dist directory
+    twine upload dist/* --username $PYPI_USERNAME --password $PYPI_PASSWORD  # publish
+else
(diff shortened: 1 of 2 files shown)


deploy-if-version-bump.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -e
+if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+    git tag `python setup.py --version`
+    git push --tags  # update the repository version
+    python setup.py bdist_wheel  # build this package in the dist directory
+    twine upload dist/* --username $PYPI_USERNAME --password $PYPI_PASSWORD  # publish
+else
+    echo "No deployment - Existing release with `python setup.py --version` version number"
```
fpagnoux: > As an OpenFisca developper, it is not clear to me how `Existing release with 16.1.0 version number` is related to `No deployment`. The fact that a version already exists is the _technical_ criterion, not the _functional_ reason.
> What about something like : `No deployment - Only non-functional elements were modified in this change` ?

PR #27, opened 2017-12-15
Improve Enums performance

fpagnoux:
Connected to openfisca/openfisca-france#831

Adapt country template to https://github.com/openfisca/openfisca-core/pull/605

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## 2.0.0 - [#24](https://github.com/openfisca/country-template/pull/24)
+## 2.0.1 - [#24](https://github.com/openfisca/country-template/pull/24) [#27](https://github.com/openfisca/country-template/pull/27)
 
-#### breaking change
+_Note: the 2.0.0 version has been unpublished due to performance issues_
+
+#### Breaking change
 
 * Details:
   - Upgrade to Core v21
@@ -69,11 +71,9 @@ name: Household living in a 40 sq.meters accomodation while being free lodgers
 
 #### Python API
 
-When calculating an enum variable in Python, the output will be an array of Enum items.
+When calculating an enum variable in Python, the output will be an [EnumArray](http://openfisca.readthedocs.io/en/latest/enum_array.html#module-openfisca_core.indexed_enums).
 
-> Each Enum item has:
-> - a `name` property that contains its key (e.g. `tenant`)
-> - a `value` property that contains its description (e.g. `"Tenant or lodger who pays a monthly rent"`)
+See more on the OpenFisca-Core [changelog](https://github.com/openfisca/openfisca-core/blob/enums-perfs/CHANGELOG.md#2102-589-600-605).
 
 ## 1.4.0 - [#26](https://github.com/openfisca/country-template/pull/26)
 
--- a/openfisca_country_template/variables/housing.py
+++ b/openfisca_country_template/variables/housing.py
@@ -8,7 +8,6 @@
 from openfisca_core.model_api import *
 # Import the entities specifically defined for this tax and benefit system
 from openfisca_country_template.entities import *
-from enum import Enum
 
 
 # This variable is a pure input: it doesn't have a formula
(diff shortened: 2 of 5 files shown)


openfisca_country_template/variables/housing.py
```diff
@@ -27,17 +27,9 @@ class rent(Variable):
...
-    tenant = u'Tenant'
-    owner = u'Owner'
-    free_lodger = u'Free logder'
-    homeless = u'Homeless'
-
-
 class housing_occupancy_status(Variable):
     value_type = Enum
-    possible_values = HousingOccupancyStatus
+    possible_values = HousingOccupancyStatus  # HousingOccupancyStatus is defined in base.py
```
Anna-Livia: Why do we have to put all Enums in the base.py file ?
I know that is the way we did it for openfisca-France but I thought it was an idiosynchrasy of that package.

CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
...
-## 2.0.0 - [#24](https://github.com/openfisca/country-template/pull/24)
+## 2.0.1 - [#24](https://github.com/openfisca/country-template/pull/24) [#27](https://github.com/openfisca/country-template/pull/27)
 
-#### breaking change
+_Note: the 2.0.0 version has been unpublished due to performance issues_
+
+#### Breaking change
 
 * Details:
   - Upgrade to Core v21
```
Anna-Livia: > Add the information that Enums are in the base.py file ?

openfisca_country_template/variables/housing.py
fpagnoux: I ran into a very stange bug when running core tests. After investigating, I found more deep issues:

[This import](https://github.com/openfisca/country-template/blob/master/openfisca_country_template/variables/taxes.py#L12) worked as expected by pure luck. It is not actually a relative import, but an absolute one. Relative imports are not allowed in our context, as the variables files are loaded one by one independently. It turns out the `housing.py` file was parsed first. Then the python importer recorded that it has loaded a module named `housing`. When it parses `taxes.py`, it sees `from housing import ...`. It then used the recorded module. If `taxes.py` was read before `housing.py`, the import would fail 😱.

To avoid this issue, the first thing I tried was to replace `from housing import HousingOccupancyStatus
` by `from openfisca_country_template.variables.housing import HousingOccupancyStatus`. With this the import works in all case, but another issue appears.

In the fast indexed enums, we rely on a [comparison of enums](https://github.com/openfisca/openfisca-core/blob/enums-perfs/openfisca_core/indexed_enums.py#L75) when comparing an `EnumArray` with an enum item. This only works if the two compared enums are _exactly_ the same object.

With the absolute import, we ended up importing `housing` twice: once when we parse `housing.py`, and another time when we parse `taxes.py` and run into the absolute import. Usually Python deals well with imports, but this doesn't work as we manually import modules from files. The double import make the comparison mentioned in the last paragraph to fail : the two enums compared have been imported separatly, they are not the same object.

Solutions i thought about:

- Make the enum comparison more robust. I didn't find an easy way to do it, and there is a risk to degrade the performances, and we know this part of the code must be fast.
- Fixing the way we import the variable files to avoid double imports. If we record the module`housing.py` as `openfisca_country_template.variables.housing` instead of just `housing`, the double import could be avoided. But I don't see any clean and robust way of doing it.
- Moving all the enums definitions to `base.py`. But it seems pretty heavy and not so  natural. I've reverted it.
- The new solution I implemented, which is avoiding any import by accessing the enum through the `EnumArray`. 

The last one seemed less heavy and more natural to me, but a different opinion on this is very much welcome !

Anna-Livia:
The tests work well locally but not on circleci. Maybe update the circle.yaml fil to DL the enum-perf branch of openfisca-core could work ?

openfisca_country_template/variables/housing.py
Anna-Livia: Sounds great, I will see if that change can be applied to openfisca-France :)

maukoquiroga:
@fpagnoux Do you have the performance delta at hand ?

fpagnoux:
@maukoquiroga check out https://github.com/openfisca/openfisca-core/pull/605

PR #31, opened 2018-03-21
Introduce python script for bootstrapping

corrado-petrelli:
We tested this script on Windows 10, Ubuntu and Debian.
This script will substitute the bash script on README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,27 +4,7 @@ This repository is here to help you quickly bootstrap and use your own OpenFisca
 
 ## Bootstrapping your Country Package
 
-This set of instructions will create your own copy of this boilerplate directory and customise it to the country you want to work on:
-
-```sh
-COUNTRY_NAME=France  # set the name of your country here; you should keep all capitals, and replace any spaces in the name by underscores
-URL=https://github.com/openfisca/openfisca-france  # set here the URL of the repository where you will publish your code.
-
-lowercase_country_name=$(echo $COUNTRY_NAME | tr '[:upper:]' '[:lower:]')
-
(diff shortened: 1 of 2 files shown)


MattiSG:
Thanks, that's a very ambitious PR!

I'm afraid it might be a bit too much for the scope of the country template, and makes assumptions that we can not be certain of.

For example, shipping a list of all existing countries is a strong limitation: some of the “country packages” are not necessarily _countries_ in the list of nation states: the township of Barcelona has modelled [their tax and benefit system](https://github.com/jvalduvieco/openfisca-barcelona) from this country template, and would not have been able to start with this user interface.

In a similar fashion, the New Zealand instance is named “[Aotearoa](https://github.com/ServiceInnovationLab/openfisca-aotearoa)”, after the Maōri name of the country. It is impossible to define these elements beforehand, and forcing reusers to go through this limited list is a very constraining.

Adding a graphical user interface is also quite a burden for maintenance. We would need to test this in all supported environments (Windows, macOS, GNU/Linux), and I am not confident we have the resources for such an investment.

Thank you again for your contribution. I will unfortunately have to go forward and close this PR.
If someone else from @openfisca/core-contributors wants to have another look and maybe help you go forward with automated GUI tests and adding free-form fields, they can reopen this as such a feature would definitely add a more professional feeling to OpenFisca!
In its current state, however, it is unfortunately a bit beyond what I believe we can maintain 🙂

corrado-petrelli:
No problem, I and @LorenzoStacchioDev just wanted to help!

PR #32, opened 2018-04-06
Improve spelling

verbman:
* Minor change.
* Impacted periods: all.
* Impacted areas: `variables/housing`
* Details:
  - Improved Spelling

- - - -

These changes _(remove lines which are unrelevant to your contribution)_:
- Change non-functional parts of this repository (for instance editing the README)

PR #34, opened 2018-04-06
Improve spelling

MattiSG:
Supersedes #32.

#### Breaking change

* Impacted periods: all.
* Impacted areas: `variables/housing`
* Details:
  - Fix spelling by renaming `accomodation_size` variable to `accommodation_size`

#### Other changes

* Minor change.
* Impacted areas: no functional impact.
* Details:
  - Improved spelling

- - - -

These changes:

- Impact the OpenFisca-Country-Template public API (for instance renaming or removing a variable)
- Change non-functional parts of this repository (for instance editing the README)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 2.1.1 - [#34](https://github.com/openfisca/country-template/pull/34)
+
+* Minor change.
+* Impacted areas: no functional impact.
+* Details:
+  - Improved spelling
+
 ## 2.1.0 - [#29](https://github.com/openfisca/country-template/pull/29) [#30](https://github.com/openfisca/country-template/pull/30)
 
 * Tax and benefit system evolution
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,11 @@
 ## 2.1.0 - [#29](https://github.com/openfisca/country-template/pull/29) [#30](https://github.com/openfisca/country-template/pull/30)
 
 * Tax and benefit system evolution
-* Impacted areas: 
+* Impacted areas:
   - Parameters `general`
   - Variables `benefits`
 * Details:
-  - Add a parameter and a variable with non ascii characters 
+  - Add a parameter and a variable with non ascii characters
     - Introduce `age_of_retirement` parameter
     - Introduce `pension` variable
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Changelog
 
-### 2.1.1 - [#34](https://github.com/openfisca/country-template/pull/34)
+# 3.0.0 - [#34](https://github.com/openfisca/country-template/pull/34)
+
+#### Breaking change
+
+* Tax and benefit system evolution.
+* Impacted periods: all.
+* Impacted areas: `variables/housing`
+* Details:
+  - Fix spelling by renaming `accomodation_size` variable to `accommodation_size`
+
+#### Other changes
 
 * Minor change.
 * Impacted areas: no functional impact.
@@ -63,10 +73,10 @@ Now:
 Before:
 
 ```
-name: Household living in a 40 sq. metres accomodation while being free lodgers
+name: Household living in a 40 sq. metres accommodation while being free lodgers
   period: 2017
   input_variables:
-    accomodation_size:
+    accommodation_size:
       2017-01: 40
     housing_occupancy_status:
       2017-01: 2
@@ -76,10 +86,10 @@ name: Household living in a 40 sq. metres accomodation while being free lodgers
 Now:
 
 ```
-name: Household living in a 40 sq. metres accomodation while being free lodgers
+name: Household living in a 40 sq. metres accommodation while being free lodgers
   period: 2017
   input_variables:
-    accomodation_size:
+    accommodation_size:
       2017-01: 40
     housing_occupancy_status:
       2017-01: free_lodger
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ Now:
 Before:
 
 ```
-name: Household living in a 40 sq.meters accomodation while being free lodgers
+name: Household living in a 40 sq. metres accomodation while being free lodgers
   period: 2017
   input_variables:
     accomodation_size:
@@ -69,7 +69,7 @@ name: Household living in a 40 sq.meters accomodation while being free lodgers
 Now:
 
 ```
-name: Household living in a 40 sq.meters accomodation while being free lodgers
+name: Household living in a 40 sq. metres accomodation while being free lodgers
   period: 2017
   input_variables:
     accomodation_size:
--- a/openfisca_country_template/parameters/general/age_of_retirement.yaml
+++ b/openfisca_country_template/parameters/general/age_of_retirement.yaml
@@ -1,5 +1,5 @@
 # A parameter can contain utf-8 characters.
-description: Age of retirement (in years). Âge de la retraite (en années).
+description: Age of retirement (in years). In French âge de la retraite (en années).
 reference: https://fr.wikipedia.org/wiki/Retraite_(économie)
 values:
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -56,7 +63,7 @@ Now:
 Before:
 
 ```
-name: Household living in a 40 sq.meters accomodation while being free lodgers
+name: Household living in a 40 sq. metres accomodation while being free lodgers
```
Anna-Livia: Seems to me `meter`is the US spelling and `metre` is the british-inspired spelling. I don't know what the rule is on the version of English to use :/

CHANGELOG.md
```diff
@@ -69,7 +76,7 @@ name: Household living in a 40 sq.meters accomodation while being free lodgers
 Now:
 
 ```
-name: Household living in a 40 sq.meters accomodation while being free lodgers
+name: Household living in a 40 sq. metres accomodation while being free lodgers
```
Anna-Livia: same

openfisca_country_template/parameters/general/age_of_retirement.yaml
```diff
@@ -1,5 +1,5 @@
 # A parameter can contain utf-8 characters.
-description: Age of retirement (in years). Âge de la retraite (en années).
+description: Age of retirement (in years). In French âge de la retraite (en années).
```
Anna-Livia: Maybe we could add ticks, quotes or parenteses around the French translation ?
And I'm wondering if `retirement age` woundn't be better than Age of retirement ?

openfisca_country_template/tests/housing_tax.yaml
```diff
@@ -1,7 +1,7 @@
 # Test files describe situations and their expected outcomes
 # We can run this test on our command line using `openfisca-run-test housing_tax.yaml`
 
-- name: Tenant living in a 40 sq.meters accomodation
+- name: Tenant living in a 40 sq. metres accomodation
```
Anna-Livia: see above

openfisca_country_template/tests/housing_tax.yaml
```diff
@@ -11,7 +11,7 @@
   output_variables:
     housing_tax: 400
 
-- name: Free_lodgers living in a 40 sq.meters accomodation
+- name: Free_lodgers living in a 40 sq. metres accomodation
```
Anna-Livia: same

openfisca_country_template/tests/housing_tax.yaml
```diff
@@ -21,7 +21,7 @@
   output_variables:
     housing_tax: 0
 
-- name: Household living in a 100 sq.meters accomodation
+- name: Household living in a 100 sq. metres accomodation
```
Anna-Livia: same

openfisca_country_template/variables/benefits.py
```diff
@@ -34,7 +34,7 @@ class housing_allowance(Variable):
     value_type = float
     entity = Household
     definition_period = MONTH
-    label = "Housing allowange"
+    label = "Housing allowance"
```
Anna-Livia: 🙌

openfisca_country_template/tests/housing_tax.yaml
Anna-Livia: Should be `accommodation`

openfisca_country_template/tests/housing_tax.yaml
Anna-Livia: Should be `accommodation`

openfisca_country_template/tests/housing_tax.yaml
Anna-Livia: Should be `accommodation`

CHANGELOG.md
MattiSG: Indeed, as I was told by @Br3nda, “`meter` is the US spelling, `metre` is everyone's else. A meter is the tool for measurement, a metre is the unit of distance.” :wink: That was actually the point of the initial commit from @verbman. The fact that native speakers feel compelled to open a PR for that is enough to justify the change for me.

openfisca_country_template/parameters/general/age_of_retirement.yaml
MattiSG: The simple fact that we use French to demonstrate UTF-8 is deficient. If we are to spend more time on that I'd rather have us find a way to bring in UTF-8 descriptions that make sense in English.
For me, that's already an improvement over the previous phrasing and I don't feel like doing a lot more.

maukoquiroga: > And I'm wondering if retirement age woundn't be better than Age of retirement ?

Seems to me `retirement age` is the correct spelling, at least according to [Wikipedia](https://en.wikipedia.org/wiki/Retirement_age)

maukoquiroga: > The simple fact that we use French to demonstrate UTF-8 is deficient.

Sure. I propose though we keep consistency with https://github.com/openfisca/country-template/pull/34/files#diff-e12bf04429356106f9750727cc8aea46R57, and use _In French: âge de la retraite (en années)._

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name='OpenFisca-Country-Template',
-    version='2.1.0',
+    version='2.1.1',
```
maukoquiroga: Bump it up 😃

fpagnoux:
Careful, following this PR, this repo's `master` is not compatible with Core, and make the test crash.

It would be vicious to postpone the adaptation and let it to the next core developper who will try to update the country template and wonder why some tests are failing without connection with their work 😅

maukoquiroga:
I didn't realise the testing circular dependency until yesterday (that'll render moving from Python 2 to 3 a rather complicated matter, for example).

How could we nudge contributors? Check to core version in setup?

That being said, the choice is rather awkward. Is there a reason to have not just done unit testing in the core ?

MattiSG:
Huh? I'm not sure I understand. How could this PR break compatibility with Core? And if so, why didn't the CI catch it? 🤔

fpagnoux:
> I didn't realise the testing circular dependency until yesterday (that'll render moving from Python 2 to 3 a rather complicated matter, for example).

This won't be the first time we need to change both Core and Country-Template at the same time. We usually solve this by manually releasing a `rc0` version of country-template, and using [this kind of dependency](https://github.com/openfisca/openfisca-core/blob/20.0.0/setup.py#L35) for core.

> How could we nudge contributors? Check core's version in setup.py?

I'd say this issue is not frequent enough to think to hard about it. Country-template is a core package that is mainly edited by the core team.

> That being said, the choice is rather awkward, isn't it? Why not just unit testing the core ?

To properly test the core, you'd have to build a small model of a country anyway. The choice has been made to make the _testing_ country and the _template_ country the same. The advantage is that it guarantees that the country template stays up to date.

> Huh? I'm not sure I understand. How could this PR break compatibility with Core? And if so, why didn't the CI catch it? 🤔

Sorry, I was not really clear: this PR didn't break the _real dependency_ of `Country-Template` towards `Core`, it broke the _test dependency_ of `Core` towards `Country-Template`. 

More details:

- This PR is a breaking change for `Country-Template`, because some variables were renamed.
- Core uses `Country-Template` as a test dependency. 
- Before Mauko's update, the `master` version of Core did not past its tests with the new version of the country template, because some tests were looking for a variable using the old name.
- The CI didn't raise anything because the CI doesn't run Core tests: it runs Country Templates tests.

It's not a huge deal, just a potential surprise for the next person developping core and wondering why are the tests are not passing.

MattiSG:
Thanks a lot @fpagnoux, very clear  :+1:

To avoid breaking Core, its dependency to Template should be more strict ( >=2 <3 in the current case).

Then, we could add a test in the Template CI that checks if the Core dependency to Template matches the one in the current branch. Here, this test would have failed as soon as the breaking change was introduced, forcing us to fix the problem before it goes live  :)

PR #40, opened 2018-05-08
Migrate to CircleCI 2.0

MattiSG:
CircleCI 1.0 is being phased out. By the end of August 2018, all projects based on the Circle 1.0 syntax won't build. This changeset updates the definition to the 2.0 syntax, and should bring substantial speed improvements, at the expense of an IMO [much less readable](https://discuss.circleci.com/t/2-0-config-syntax-is-ugly/22225) file.

I also decided to move the deployment scripts to the now mandatory `.circleci` directory so as to remove the root folder clutter, as I don't believe one would (nor should) deploy manually from their machines with these scripts. If that's a problem to anyone we can drop that separate commit.

Bumping the version number is not necessary as this is a non-functional change, but it might be useful to do it still, so that reusers can know that their installation will be compatible with Circle 2.0 if they based their fork from some specific version number on.

- - - -

* Technical improvement.
* Impacted periods: none.
* Impacted areas: continuous integration.
* Details:
  - Activate CircleCI 2.0

These changes:

- Change non-functional parts of this repository (for instance editing the README)


fpagnoux:
Arg, indeed, it seems that we have been working on really similar topics. 

I like the concision and the clarity of this config more than the current one in Core.

I must say I'm not a big fan of sodoing instead of using a virtualenv. Not using virtualenvs is always a source of problems.

I'll try to merge the configs to get something concise that work for Python 2 and 3.

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+          name: Run tests
+          command: make test
+
+      - run:
+          name: Check version number has been properly updated
+          command: |
+            git fetch
+            ./.circleci/check-version-bump.sh
+
+  deploy:
```
fpagnoux: Have you really tried this ? 

It seems like CircleCI doesn't persist anything from one job to another, unless you use [workflows or cache](https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/). 

So when running this job, I'd expect the current folder not to be checked out, and the dependencies not being installed.

MattiSG: I did only partially: the syntax was correct but I was missing a username and password to actually test the deployment. I agree with your assumptions of the build folder being empty.

MattiSG:
Superseded by #41.

PR #41, opened 2018-05-15
Publish Python 3 compatible packages and migrate to CircleCI 2.0

fpagnoux:
Relates to openfisca/openfisca-core#630

Deployment works for[ Python 2](https://circleci.com/gh/openfisca/country-template/283) and [Python 3](https://circleci.com/gh/openfisca/country-template/282)


deploy-if-version-bump.sh
```diff
@@ -1,11 +1,20 @@
...
+      git tag `python setup.py --version`
+      git push --tags  # update the repository version
+      python setup.py bdist_wheel  # build this package in the dist directory
+      twine upload dist/* --username $PYPI_USERNAME --password $PYPI_PASSWORD  # publish
+  else
+      echo "No deployment - Only non-functional elements were modified in this change"
+  fi
+fi
+
+if [ $PY_VERSION = "3" ]; then
```
MattiSG: So if the Python version is `3` then there is no functional changes check. Is that really the goal?

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name='OpenFisca-Country-Template',
-    version='3.0.2',
+    version='3.0.3',
```
MattiSG: Changelog and setup versions do not match.

> I prefer the minor bump over the patch bump 🙂

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+    <<: *deploy
+    docker:
+      - image: python:3.6
+
+workflows:
+  version: 2
+  country-template:
+    jobs:
+      - checkout
+      - dependencies:
```
MattiSG: > What about suffixing with `_python2` like it is done with `_python3`?

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+            - checkout
+      - tests_python3:
+          context: Py3
+          requires:
+            - dependencies_python3
+      - deploy:
+          context: Py2
+          requires:
+            - tests
+            - tests_python3
```
MattiSG: This reads like “if I'm in `Py2` context, run `tests_python3`” 🤔 Is this normal?

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+          filters:
+            branches:
+              only: master
+      - deploy_python3:
+          context: Py3
+          requires:
+            - tests
```
MattiSG: This reads like “if I'm in `Py3` context, run `tests` (which is Python2)” 🤔 Is this normal?

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+
+      - restore_cache:
+          keys:
+            - v1-{{ .Environment.PY_VERSION }}-dependencies-{{ .Environment.CIRCLE_SHA1 }}
+
+      - run:
+          name: Deploy (if version bump)
+          command: |
+            . /tmp/venv/country-template/bin/activate
+            . deploy-if-version-bump.sh
```
MattiSG: Unless CircleCI does something special, the `.` is a non-readable equivalent of `source`. I recommend we use `source` over `.` for readability. I also believe that `. deploy-if-version-bump.sh` should not be sourced, but directly executed, and that it currently works only as a side-effect of sourcing.

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+    docker:
+      - image: python:2.7.14
+    environment:
+      PYPI_USERNAME: openfisca-bot
+      # PYPI_PASSWORD: this value is set in CircleCI's web interface; do not set it here, it is a secret!
+
+    steps:
+      - restore_cache:
+          keys:
+            - v1-checkout-{{ .Environment.CIRCLE_SHA1 }}
```
MattiSG: Such a key prevents the cache from having almost any use: only a rerun with cache would benefit from it.

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+            make test
+            . check-version-bump.sh
+
+  tests_python3:
+    <<: *tests
+    docker:
+      - image: python:3.6
+
+  deploy: &deploy
+    working_directory: ~/country-template
```
MattiSG: > I don't think you need to state this explicitly.

.circleci/config.yml
```diff
@@ -0,0 +1,158 @@
...
+
+      # Uncomment and adapt the next line to use a particular feature branch of OpenFisca-Core to run Circle CI tests
+      # - run:
+      #     name: Install a particular feature branch of OpenFisca-Core
+      #     command: |
+      #       . /tmp/venv/country-template/bin/activate
+      #       pip install --editable git+https://github.com/openfisca/openfisca-core.git@BRANCH_NAME#egg=OpenFisca-Core
+
+      - save_cache:
+          # PY_VERSION is set in the context
```
MattiSG: “the context” is ambiguous. I personally don't understand where this is coming from.

MattiSG:
I must say that I am a bit surprised and frustrated that my PR on CircleCI 2.0 migration (https://github.com/openfisca/country-template/pull/40) has been lingering for over a week, and this one does essentially the same thing. I had already invested in this long, error-prone process, and now we have two very different implementations.

@fpagnoux I'll let you have a look at both implementations and assess if it would be easier to add Python3 support on top of #40 or integrate #40's syntax refinements in this PR.

How can we avoid such collaboration mishaps in the future?

fpagnoux:
I updated the configuration to look more #40's one, which I find simpler.

To handle both Python 2 and Python 3, I went for duplication over complexity.
- Reusing configuration means using complex YAML features that not many know about
- Making one cache key by python version is tricky without complex and non straightforward console configuration (contexts) 

@MattiSG let me know if that still seem too complicated for a package that aims to be  cloned and re-used by many openfisca contributors.

setup.py
fpagnoux: Done

.circleci/config.yml
fpagnoux: Done

.circleci/config.yml
fpagnoux: Got rid of contexts

.circleci/config.yml
fpagnoux: Got rid of contexts

.circleci/config.yml
fpagnoux: Done

.circleci/config.yml
fpagnoux: I think it was only used to persist the files from one job to another. But yes, checksuming setup.py seems better.

.circleci/config.yml
fpagnoux: Indeed we don't

.circleci/config.yml
fpagnoux: Got rid of that

.circleci/deploy-python-3.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -e
+
+VERSION_CHANGE_TRIGGERS="setup.py MANIFEST.in openfisca_core openfisca_web_api_preview"
```
fpagnoux: @MattiSG I'm not sure about repeating this in 3 different files, as the whole continuous publishing relies on this variable to have the same values in the 3 files. 

However, I didn't manage to share it. I tried to declare it in a separate shell file and to source it from each of the 3 scripts, but the `source` would not find the file (I think because relative paths in a script are always relative to where you launch the script, not where the script is).

If you know how to make it better, please proceed 🙂

.circleci/deploy-python-2.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+VERSION_CHANGE_TRIGGERS="setup.py MANIFEST.in openfisca_core openfisca_web_api_preview"
+
+if git diff-index --quiet HEAD^ -- $VERSION_CHANGE_TRIGGERS ":(exclude)*.md"
```
MattiSG: We're only checking for changes in the last commit, not in the whole branch. Is that expected?

.circleci/deploy-python-3.sh
MattiSG: I'll just extract that in a new script :wink:

If you're curious, [this is how](https://github.com/MattiSG/DotFiles/blob/1e33187131229a23aa2467f6f54e44d800c3a470/source/prompt.sh#L1) I usually solve the problem.

setup.py
```diff
@@ -16,7 +16,7 @@
     url='https://github.com/openfisca/openfisca-country-template',
     include_package_data = True,  # Will read MANIFEST.in
     install_requires=[
-        'OpenFisca-Core >= 23, < 24.0',
+        'OpenFisca-Core >= 23.1, < 24.0',
```
MattiSG: > Any reason for that change here or is it just a knight? 😛

.circleci/config.yml
```diff
@@ -0,0 +1,139 @@
...
+
+      - run:
+          name: Run tests
+          command: make test
+
+      - run:
+          name: Check version number has been properly updated
+          command: |
+            git fetch
+            ./.circleci/check-version-bump.sh
```
MattiSG: > No need for the `./` when you're already giving a directory path.

.circleci/config.yml
```diff
@@ -0,0 +1,139 @@
...
+
+  deploy_python2:
+    docker:
+      - image: python:2.7.14
+    environment:
+      PYPI_USERNAME: openfisca-bot
```
MattiSG: Since this is a template, there should be a comment here about customising this value.

.circleci/config.yml
```diff
@@ -0,0 +1,139 @@
...
+            ./.circleci/deploy-python-3.sh
+
+workflows:
+  version: 2
+  build_and_deploy:
+    jobs:
+      - build_python2
+      - build_python3
+      - deploy_python2:
+          requires:  # Only deploy if tests passed both with Python 2 and Python 3
```
MattiSG: > This is quite readable in the config. This comment would add more value by explaining _why_ we don't want to deploy unless both versions pass.

.circleci/deploy-python-2.sh
```diff
@@ -1,11 +1,14 @@
 #!/bin/sh
 
 set -e
-if ! git rev-parse `python setup.py --version` 2>/dev/null ; then
+
+if ./detect-functional-changes.sh
+then
     git tag `python setup.py --version`
     git push --tags  # update the repository version
```
MattiSG: I don't see the difference between Python2 and Python3 deployment beyond this initial tagging.
If indeed there is none, then I think the tagging should be done in CI (or in a script), and that the deployment should be a single script as it was. Right now, there is an apparent dependency of deployment of Python3 to deployment of Python2, which is both absurd and not guaranteed by CI; there is also unnecessary code duplication in the deployment scripts.

.circleci/deploy-python-2.sh
```diff
@@ -1,11 +1,14 @@
...
     python setup.py bdist_wheel  # build this package in the dist directory
     twine upload dist/* --username $PYPI_USERNAME --password $PYPI_PASSWORD  # publish
+    ssh -o StrictHostKeyChecking=no deploy-api@fr.openfisca.org  # Deploy the OpenFisca-France public API
```
MattiSG: I don't understand why this is in the country template 🤔

.circleci/detect-functional-changes.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+VERSION_CHANGE_TRIGGERS="setup.py MANIFEST.in openfisca_country_template"
+
+if git diff-index --quiet origin/master -- $VERSION_CHANGE_TRIGGERS ":(exclude)*.md"
```
MattiSG: @fpagnoux Beware, this is not the same check as it was before I factored it: please double-check based on https://github.com/openfisca/country-template/pull/41#discussion_r188825589 that this is the proper check we want to apply here!

setup.py
fpagnoux: The changelog claims that we are making this package compatible with python 3.
We thus need to make sure we request a core version that is compatible with python 3. Actually, 23.1.1 may be even better.

.circleci/config.yml
fpagnoux: Ok, I thought it was unclear from a previous comment. Do you think we should not require the `build_python2` for  `deploy_python3`? I guess that we are on master, so technically the tests have already passed...

.circleci/deploy-python-2.sh
fpagnoux: Excessive copy pasting

.circleci/deploy-python-2.sh
fpagnoux: There is a subtlety here.

We are in the context of deployment. So there is not more feature branch involved: we are on `master`, and want to know if the new version of master brings any functional change compared to the last version of master. Hence the comparison with `HEAD^`. 

This is different when we are testing a feature branch, and running `check-version-bump.sh`. In that case, we indeed want to compare our branch to the current version of master.

So we can't really share the whole `detect-functional-changes.sh` as is. I'll try to find a solution though.

fpagnoux:
Tests ran:
- A non functional change merged on master does not trigger a deployment: [py2](https://circleci.com/gh/openfisca/country-template/317) [py3](https://circleci.com/gh/openfisca/country-template/318)
- A functional change merged on master does trigger a deployment: [py2](https://circleci.com/gh/openfisca/country-template/322) [py3](https://circleci.com/gh/openfisca/country-template/321)
- A feature branch with no functional change does not need to update its version number: [py2](https://circleci.com/gh/openfisca/country-template/328) [py3](https://circleci.com/gh/openfisca/country-template/327)
- A feature branch with functional change that does not update the version number does not pass the tests: [py2](https://circleci.com/gh/openfisca/country-template/330) [py3](https://circleci.com/gh/openfisca/country-template/329)
- A feature branch with functional change that does not update the changelog does not pass the tests: [py2](https://circleci.com/gh/openfisca/country-template/331) [py3](https://circleci.com/gh/openfisca/country-template/332)
- A feature branch with functional change that updates the version number and the changelog does pass the tests: [py2](https://circleci.com/gh/openfisca/country-template/333) [py3](https://circleci.com/gh/openfisca/country-template/334)

Now let's please get done with this 😓

.circleci/config.yml
fpagnoux: Done

.circleci/config.yml
fpagnoux: I think you fixed it

.circleci/config.yml
fpagnoux: Removed the dependency, as it's not really necessary

.circleci/deploy-python-2.sh
fpagnoux: Done

.circleci/detect-functional-changes.sh
fpagnoux: See reply on the said comment

.circleci/config.yml
MattiSG: I thought it would be because we don't really want to publish a new version for only one version: I wouldn't want to deal with version `23.1.2` existing in Python3 but not in Python2 and the other way around for `23.1.3` 😅

.circleci/config.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-py2-{{ checksum "setup.py" }}
+
+      - run:
+          name: Check for functional changes
+          command: if .circleci/detect-functional-changes.sh ; then circleci step halt ; fi
```
MattiSG: Why don't we just `.circleci/detect-functional-changes.sh`? Shouldn't deployment fail?

.circleci/config.yml
```diff
@@ -0,0 +1,149 @@
...
+
+workflows:
+  version: 2
+  build_and_deploy:
+    jobs:
+      - build_python2
+      - build_python3
+      - deploy_python2:
+          requires:
+            - build_python2
```
MattiSG: I don't think we want to publish a Python2 package that does not pass Python3 builds. How would we manage a `23.1.2` tag that got published in Python2 but failed in Python3? We would then have to publish a `23.1.3` tag that would be available in both environments… So all version numbers could co-exist differently in each language version 😕

.circleci/detect-functional-changes.sh
```diff
@@ -0,0 +1,13 @@
+#! /usr/bin/env bash
+
+VERSION_CHANGE_TRIGGERS="setup.py MANIFEST.in openfisca_country_template"
+
+if [[ "$CIRCLE_BRANCH" == "master" ]]
```
MattiSG: > No need for quotes there :wink:

.circleci/publish-git-tag.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/env bash
```
MattiSG: > When you're not using bash-specific instructions, `sh` is safer: `bash` is an implementation with extensions while `sh` is an interface.

.circleci/config.yml
MattiSG: Alright I understood. No, it should not fail, it should be green but not do anything. Gotcha.

MattiSG:
> Tests ran: […]

I reviewed all links and agree with all the tests. Thank you so much for this list of links, it has proven extremely useful. It is also very reassuring to see how all possible cases have been covered 👏

.circleci/publish-git-tag.sh
maukoquiroga: I'd add that _usually_ `sh` is just a pointer to `bash`, but *not always*.

.circleci/detect-functional-changes.sh
fpagnoux: ok

.circleci/config.yml
fpagnoux: OK, I misinterpreted a previous comment.

.circleci/detect-functional-changes.sh
```diff
@@ -2,12 +2,9 @@
 
 VERSION_CHANGE_TRIGGERS="setup.py MANIFEST.in openfisca_country_template"
 
-if [[ "$CIRCLE_BRANCH" == "master" ]]
-then LAST_MASTER_VERSION="HEAD^"
-else LAST_MASTER_VERSION="origin/master"
-fi
+last_tagged_commit=`git describe --tags --abbrev=0 --first-parent`  # --first-parent ensures we don't follow tags not published in master through an unlikely intermediary merge commit
```
fpagnoux: Not sure I fully understand the motive of the change.

Can you give a concrete example of a situation where the previous implem would not do its job?

One risk I can see here is if for some reason a tag is published in the middle of the feature branch. Then the script would not compare the feature branch to the last published version, but to this tag.

fpagnoux:
Changed the workflow so that we deploy *only* if the two builds have passed.

.circleci/detect-functional-changes.sh
MattiSG: This change has the technical benefit of working in both cases without needing specific logic based on the branch name.

In terms of logic, it makes the assumption that all tagged commits are published. The previous implementation made the assumption that all commits on `master` are published. I am more confident that no accidental tagging will happen than no commits on `master` can accidentally happen.
In the same way, if we want to unpublish a version (or mark a version as unpublished), we now simply have to remove a tag. Removing a commit on `master` is harder :wink:

If you want added rigidity (which would go beyond what was already the case), I'm happy to add a check for the tag to be `x.y.z`, preventing preversions from being taken into account 🙂

MattiSG:
I'm changing the required CI checks to allow for merging.

MattiSG:
Done.

.circleci/detect-functional-changes.sh
fpagnoux: Ok, fair enough.

I was thinking that under our workflow, accidental commits on master are pretty rare, if not inexistent, but we can't assume that it will be the same for all re-users.

Let's merge this as-is, and see if we need more refinement later.

fpagnoux:
Pew, that was painful 😅

.circleci/detect-functional-changes.sh
MattiSG: > under our workflow, accidental commits on master are pretty rare

Totally agree! It just happens that I believe the semantics of “this version was published” are more closely aligned with “this version is tagged” than “this version was merged into `master`”. Thanks to all the automation we've made, this is luckily all aligned, but I felt this implementation was both clearer and shorter 😉

maukoquiroga:
Wonderful thanks !

PR #42, opened 2018-05-19
Improve installation procedure

MattiSG:
The current instructions recommend to clone the country-template. However, that means inheriting a whole history and, even more annoyingly, tags. These version tags will conflict with the actual country package version tags unless they build upon the existing ones, which don't make sense at all as one of the first actions when publishing is to remove all the boilerplate files.

This changeset:

- Recommends to download a copy of the repository contents rather than to clone it.
- Extracts and makes more resilient the bootstrapping script.
- Simplifies some of the wording and removes redundancy in some instructions.
- Switches all openfisca.org URLs to HTTPS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### 3.1.3 - [#37](https://github.com/openfisca/country-template/pull/37)
+
+* Minor change.
+* Impacted areas: no functional impact.
+* Details:
+  - Upgrade openfisca.org references to HTTPS.
+
 ### 3.1.2 - [#38](https://github.com/openfisca/country-template/pull/38)
 
 * Add situation example using YAML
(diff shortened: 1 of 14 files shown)


MattiSG:
@openfisca/core-contributors I'm stuck without a review and, considering the lack of immediate functional impact, considering using admin powers to force merge 😕

MattiSG:
Thanks @michelbl! 😉

MattiSG:
@michelbl I just realised your “for what it's worth” is ambiguous. You've tried to execute the `bootstrap.sh` script, right? 😅

michelbl:
Heu... #47

fpagnoux:
As much as I understand the frustration of waiting 10 days without a review, and as much I appreciate @michelbl intention to help unlock the situation, I don't think this is how we should handle pull requests and reviews.

Reviewing means approving the code changes, and take responsibility for its content.

MattiSG:
Agreed @fpagnoux. I must say though that in my opinion @michelbl did take responsibility, by taking the time to fix the outcome of the merged PR by opening #47 🙂

PR #43, opened 2018-05-21
Improve age computation formula and Variables tutorial comments

MattiSG:
* Impacted periods: all.
* Impacted areas: `demographics.age`
* Details:
  - Improve reliability and accuracy of `age` formula
  - Improve variables comments

- - - -

Comments improvements are only partly a knight: PRs such as [ServiceInnovationLab/openfisca-aotearoa#15](https://github.com/ServiceInnovationLab/openfisca-aotearoa/pull/15/files#diff-e97a4b5c683152aba1c810f24751a129R12) show that reusers will copy and paste comments and that can lead to review friction.

### DIFF
--- a/openfisca_country_template/variables/benefits.py
+++ b/openfisca_country_template/variables/benefits.py
@@ -4,7 +4,7 @@
 # A variable is a property of an Entity such as a Person, a Household…
 # See https://openfisca.org/doc/variables.html
 
-# Import from openfisca-core the common python objects used to code the legislation in OpenFisca
+# Import from openfisca-core the common Python objects used to code the legislation in OpenFisca
 from openfisca_core.model_api import *
 # Import the Entities specifically defined for this tax and benefit system
 from openfisca_country_template.entities import *
--- a/openfisca_country_template/variables/benefits.py
+++ b/openfisca_country_template/variables/benefits.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
 
-# This file defines the variables of our legislation.
-# A variable is property of a person, or an entity (e.g. a household).
+# This file defines variables for the modelled legislation.
+# A variable is a property of an Entity such as a Person, a Household…
 # See https://openfisca.org/doc/variables.html
 
 # Import from openfisca-core the common python objects used to code the legislation in OpenFisca
 from openfisca_core.model_api import *
-# Import the entities specifically defined for this tax and benefit system
+# Import the Entities specifically defined for this tax and benefit system
 from openfisca_country_template.entities import *
 
 
--- a/openfisca_country_template/variables/demographics.py
+++ b/openfisca_country_template/variables/demographics.py
@@ -10,6 +10,15 @@ from openfisca_core.model_api import *
 from openfisca_country_template.entities import *
 
 
+# This variable is a pure input: it doesn't have a formula
+class birth(Variable):
+    value_type = date
+    default_value = date(1970, 1, 1)  # By default, if no value is set for a simulation, we consider the people involved in a simulation to be born on the 1st of Jan 1970.
+    entity = Person
+    label = u"Birth date"
+    definition_period = ETERNITY  # This variable cannot change over time.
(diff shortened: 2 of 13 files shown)


MattiSG:
@fpagnoux CircleCI's flake8 complaint seems a bit strange to me, since it is related to the age formula you wrote would you mind handling it? 🙂 

> ./openfisca_country_template/variables/demographics.py:12:1: F401 'numpy.datetime64' imported but unused

benjello:
Surfing on this PR: @guillett @fpagnoux : should we transpose this to france ?

openfisca_country_template/variables/benefits.py
```diff
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
 
-# This file defines the variables of our legislation.
-# A variable is property of a person, or an entity (e.g. a household).
+# This file defines variables for the modelled legislation.
+# A variable is a property of an Entity such as a Person, a Family…
```
fpagnoux: Shouldn't we use `Household`? There is no `Family` in this model.

openfisca_country_template/variables/demographics.py
```diff
@@ -1,17 +1,26 @@
...
 
-# Import from openfisca-core the common python objects used to code the legislation in OpenFisca
+# Import from openfisca-core the common Python objects used to code the legislation in OpenFisca
 from openfisca_core.model_api import *
-# Import the entities specifically defined for this tax and benefit system
+# Import the Entities specifically defined for this tax and benefit system
 from openfisca_country_template.entities import *
-
+# Import additional NumPy helpers to work on specific data types
 from numpy import datetime64
```
fpagnoux: I don't think we are actually using this import anymore, and that's why Flake is complaning.

openfisca_country_template/variables/demographics.py
```diff
@@ -21,14 +30,10 @@ class age(Variable):
     # A person's age is computed according to its birth date.
     def formula(person, period, parameters):
         birth = person('birth', period)
-        return (datetime64(period.date) - birth).astype('timedelta64[Y]')
+        birth_year = birth.astype('datetime64[Y]').astype(int) + 1970
+        birth_month = birth.astype('datetime64[M]').astype(int) % 12 + 1
+        birth_day = (birth - birth.astype('datetime64[M]') + 1).astype(int)
```
fpagnoux: > Frankly ugly, but we don't have better yet

fpagnoux:
> Surfing on this PR: @guillett @fpagnoux : should we transpose this to france ?

Probably a good idea, I think there was an issue opened about it somewhere.

openfisca_country_template/variables/demographics.py
MattiSG: I agree, but I wanted a double check from the person who wrote the initial code :wink:

openfisca_country_template/variables/demographics.py
MattiSG: I'd _really_ like that a time helper library is researched and vetted for.

MattiSG:
@fpagnoux the tests seem to show there is an issue with the formula 😕

We should probably add more tests to this in any case, including for leap years.

openfisca_country_template/variables/demographics.py
fpagnoux: We started to ask around, and there are candidates (Maya, Arrow, etc.). Though in this specific case, we need to handle _vectors_ of dates, which will be a clear limitation.
If we were to use a library, we should also probably also use it to handle periods. So IMO this is related to https://github.com/openfisca/openfisca-core/issues/670.

fpagnoux:
> @fpagnoux the tests seem to show there is an issue with the formula 😕

Indeed. I fixed a formula and added a test for leap years. Feel free to add the tests you'll find relevant 🙂.

MattiSG:
That was an annoying rebase, hopefully I got it right :wink:

MattiSG:
> hopefully I got it right

(just kidding, I double-checked that the diff from the resulting rebase to the previous branch was equal to the changes made to `master`)

PR #44, opened 2018-05-22
Return an exit code on functional changes detect

MattiSG:
Fixup #41: the deployment could never be triggered.


MattiSG:
screen shot 2018-05-22 at 14 15 27

🎉

PR #45, opened 2018-05-22
Fix instructions to run tests on core FB

fpagnoux:
Let's see how this all Circle new conf works for a non functional change!


fpagnoux:
Apparently pretty well:
image

PR #51, opened 2018-08-02
Adapt to Core v24

fpagnoux:
Connected to openfisca/openfisca-core#641
Depends on openfisca/openfisca-core#703

* Technical change
  - Make package compatible with OpenFisca Core v24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+
+### 3.3.0 - [#51](https://github.com/openfisca/country-template/pull/51)
+
+* Technical change
+  - Make package compatible with OpenFisca Core v24
+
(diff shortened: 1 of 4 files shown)


Anna-Livia:
I think that we should also change the README to indicate the need to request the extra dependencies to serve the Web API --> I made some proposition in https://github.com/openfisca/country-template/pull/52

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+
+### 3.3.0 - [#51](https://github.com/openfisca/country-template/pull/51)
```
Anna-Livia: Why not `3.2.4` ?

fpagnoux: Some argue that "making a country package compatible with a new major version of core" is more a "new feature" than a "bug fix". I have no strong opinion on the matter, but several times I've been told to use minors 🙂

PR #53, opened 2018-09-13
Add editorial content on parameter and variable

sandcha:
Connected to openfisca/legislation-explorer#124

Note : 
This branch was necessary to test openfisca/openfisca-core#717
Nevertheless, the question is: what do we want to keep as documentation example for new countries? 🙂 

- - - -

* Minor change
* Impacted areas: `parameters/benefits`, `parameters/benefits/housing_allowance` and `variables/benefits`
* Details:
  - Add `documentation` to parameters: `benefits` node and `benefits/housing_allowance`
  - Add `documentation` to `housing_allowance` variable and docstring to its formula

- - - -

These changes:

- Improve already existing parameter and variable *metadata*.

### DIFF
--- a/openfisca_country_template/parameters/benefits/index.yaml
+++ b/openfisca_country_template/parameters/benefits/index.yaml
@@ -1,3 +1,4 @@
 # This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
 
 description: Social benefits
+documentation: This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
--- a/openfisca_country_template/parameters/benefits/index.yaml
+++ b/openfisca_country_template/parameters/benefits/index.yaml
@@ -1,4 +1,9 @@
 # This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
 
 description: Social benefits
-documentation: This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
+documentation: |
+  Government support for the citizens and residents of society. 
+  They may be provided to people of any income level, as with social security, 
+  but usually it is intended to ensure that everyone can meet their basic human needs 
+  such as food and shelter.
+  (See https://en.wikipedia.org/wiki/Welfare)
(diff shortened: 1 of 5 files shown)


openfisca_country_template/variables/benefits.py
```diff
@@ -37,9 +37,15 @@ class housing_allowance(Variable):
     label = "Housing allowance"
     reference = "https://law.gov.example/housing_allowance"  # Always use the most official source
     end = '2016-11-30'  # This allowance was removed on the 1st of Dec 2016. Calculating it before this date will always return the variable default value, 0.
+    unit = 'currency-EUR'
+    documentation = '''
+    This allowance was introduced on the 1st of Jan 1980.
+    It needs the 'rent' value (same month) but doesn't care about the 'housing_occupancy_status'.
```
fpagnoux: > 👍, this kind of information adds a lot of value I think.
> I'd maybe explicit a little more what we mean by "It needs":

> To compute this allowance, the 'rent' value must be provided for the same month, but ''housing_occupancy_status'' is not necessary

openfisca_country_template/parameters/benefits/index.yaml
```diff
@@ -1,3 +1,4 @@
 # This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
 
 description: Social benefits
+documentation: This (optional) file defines metadata for the node parameters.benefits in the parameter tree.
```
fpagnoux: Same than below, a realistic example would be much better. 
Maybe a generic definition of social benefits? 

Suggestion (extracted from [this Wikipedia page](https://en.wikipedia.org/wiki/Welfare))

> Government support for the citizens and residents of society. They may be provided to people of any income level, as with social security, but usually it is intended to ensure that everyone can meet their basic human needs such as food and shelter.

openfisca_country_template/variables/benefits.py
```diff
@@ -37,9 +37,15 @@ class housing_allowance(Variable):
...
+    '''
 
     # This allowance was introduced on the 1st of Jan 1980. Calculating it before this date will always return the variable default value, 0.
     def formula_1980(household, period, parameters):
+        '''This is my specific life statement.'''
```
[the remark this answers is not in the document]
sandcha: Done. I agree on realistic examples. I simply lacked inspiration. 🙂

PR #59, opened 2018-10-19
Test against packaged version

bonjourmauko:
Child of openfisca/openfisca-france#1177

* Technical change
* Details:
  - Tests library against its packaged version
  - By doing so, we prevent some hideous bugs


pblayo (approved):
I both tested passing and failing test.

Issue #60, opened 2018-10-30
Circleci cache detects only setup.py changes


pblayo:
The problem, after merging of #59 :

```sh
openfisca-core$ make test
rm -rf build dist
find . -name '*.pyc' -exec rm \{\} \;
python -m compileall -q .
flake8 `git ls-files | grep "\.py$"`
pytest
====================================================================== test session starts =======================================================================
platform linux -- Python 3.6.7, pytest-3.9.3, py-1.7.0, pluggy-0.8.0
rootdir: /home/pblayo/openfisca-core, inifile: setup.cfg
collecting 0 items / 1 errors                                                                                                                                    
============================================================================= ERRORS =============================================================================
____________________________________________________________ ERROR collecting tests/core/test_axes.py ____________________________________________________________
tests/core/test_axes.py:5: in <module>
    from openfisca_country_template import CountryTaxBenefitSystem
../.pyenv/versions/3.6.7/envs/openfisca-core-3.6.7/lib/python3.6/site-packages/openfisca_country_template/__init__.py:8: in <module>
    from openfisca_country_template.situation_examples import couple
../.pyenv/versions/3.6.7/envs/openfisca-core-3.6.7/lib/python3.6/site-packages/openfisca_country_template/situation_examples/__init__.py:16: in <module>
    single = parse('single.json')
../.pyenv/versions/3.6.7/envs/openfisca-core-3.6.7/lib/python3.6/site-packages/openfisca_country_template/situation_examples/__init__.py:12: in parse
    with open(file_path, 'r') as file:
E   FileNotFoundError: [Errno 2] No such file or directory: '/home/pblayo/.pyenv/versions/3.6.7/envs/openfisca-core-3.6.7/lib/python3.6/site-packages/openfisca_country_template/situation_examples/single.json'
============================================================== 1 warnings, 1 error in 3.51 seconds ===============================================================
make: *** [test] Error 1
```
in .circle/config.yml : 
```
      - save_cache:
          key: v1-py2-{{ checksum "setup.py" }}
          paths:
            - /tmp/venv/country_template
            - dist
```

maukoquiroga:
Hi @pblayo, this problem _may_ propagate also to `extension-template`, and even `core` or `France`.

maukoquiroga:
Fixed by #62

PR #63, opened 2018-10-30
Cleanup CircleCI config

maukoquiroga:
Relates to #60 
Goes further than #62

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ jobs:
           key: v1-py2-{{ checksum "setup.py" }}
           paths:
             - /tmp/venv/country_template
+            - dist
 
       - save_cache:
           key: v1-py2-{{ .Revision }}
@@ -101,6 +102,7 @@ jobs:
           key: v1-py3-{{ checksum "setup.py" }}
           paths:
             - /tmp/venv/country_template
+            - dist
 
       - save_cache:
           key: v1-py3-{{ .Revision }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-py2-{{ checksum "setup.py" }}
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,9 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-py2-{{ checksum "setup.py" }}
+          keys:
+            - v1-py2-deps-{{ checksum "setup.py" }}
+            - v1-py2-build-{{ .Revision }}
```
fpagnoux: I don't think we need to restore the previous build to to run the next build

.circleci/config.yml
```diff
@@ -82,7 +83,9 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-py3-{{ checksum "setup.py" }}
+          keys:
+            - v1-py3-deps-{{ checksum "setup.py" }}
+            - v1-py3-build-{{ .Revision }}
```
fpagnoux: I don't think we need to restore the previous build to to run the next build

.circleci/config.yml
```diff
@@ -55,10 +55,9 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: v1-py2-{{ checksum "setup.py" }}
-
-      - restore_cache:
-          key: v1-py2-{{ .Revision }}
+          keys:
+            - v1-py2-deps-{{ checksum "setup.py" }}
+            - v1-py2-build-{{ .Revision }}
```
fpagnoux: Apparently, that's not how Circle works: when you provide two keys, it will checkout the second one _only_ if there is no match for the first one.  We need two separate calls.

fpagnoux:
Fixed and tested, but on a meta note, I think as a team we clearly fell in the trap of overquality on this one. #60 was good enough to fix our issues and tested.  Testing releasing is [costly](https://circleci.com/gh/openfisca/country-template/tree/improve-circle), and now we end up spending too much time for relatively superficial changes.

PR #66, opened 2018-11-28
Uniformize tests with the new API

fpagnoux:
Connected to openfisca/openfisca-core#716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,71 @@
 # Changelog
 
+### 3.6.O - [#66](https://github.com/openfisca/country-template/pull/66)
+
+* Minor change
+* Details:
+  - Adapt to OpenFisca Core v25
+  - Change the syntax of OpenFisca YAML tests
+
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+### 3.6.O - [#66](https://github.com/openfisca/country-template/pull/66)
```
maukoquiroga: Isn't it a breaking change? (I'll approve if we agree it is not, or if we bump major)

fpagnoux: Which usecase are you thinking about?

fpagnoux: I guess we could consider that a BC if we consider that users can run tests of their own that are not part of the package 🤔

Would you have time to suggest or add a few commits to explicit the breaking change?

maukoquiroga: Hey, after thinking a bit in the metro, and by the way the library is used, I guess you're right is a minor.

PR #85, opened 2020-02-10
Add `make serve-local` command to Makefile

alexsoble:
# What does this do?

Adds an additional command to the Makefile for quickly serving the API locally:

```
make serve-local
```

# Why? 

+ Makes the Makefile more comprehensive in terms of the commands it offers users.
+ Makes the command to serve the app discoverable via the Makefile.
+ Saves the user from remembering a custom-named command when getting started with a new OpenFisca project.

# Notes

+ The Makefile runs the `serve` command without any arguments as a form of quickstart; a user would need to run the full `openfisca serve` command if they want to send it command line arguments. 
+ Named the command `serve-local` under the assumption that users will want different config when serving in production.

- - - -

Minor change. These changes:

- Change non-functional parts of this repository (for instance editing the README).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+
+### 3.9.9 - [#85](https://github.com/openfisca/country-template/pull/85)
+
+* Minor change.
+* Details:
+  - Add `make serve-local` command to Makefile.
+
 ### 3.9.8 - [#83](https://github.com/openfisca/country-template/pull/83)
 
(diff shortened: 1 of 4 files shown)


alexsoble:
☝️ A short note on motivation — as I've spun up a few prototype OpenFisca projects recently, I've found a `make serve-local` shortcut useful for getting started faster. 

Opening this PR in case other newer folks would find this useful too!

maukoquiroga:
Thanks !

PR #86, opened 2020-02-11
Fix bug in bootstrap script

alexsoble:
# Notes

*  Technical improvement. 
* Details:
  - This PR substitutes in the user-supplied country name for the default country name in the project `MANIFEST.in` file when the user runs the bootstrap script.
  - If the country name is not updated in `MANIFEST.in`, the correct dependencies will not be installed by `make install` and the project will not run.
  - I tested the bootstrap script on this branch and found that `make serve-local` (new command!) built, installed, and ran the project without issues; testing the same on the master branch resulted in an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Technical change.
 * Details:
-  - Fix bootstrap script bug.
+  - Fix installation and building operations by fixing the bootstrap script.
 
 ### 3.9.9 - [#85](https://github.com/openfisca/country-template/pull/85)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 3.9.10 - [#86](https://github.com/openfisca/country-template/pull/86)
+
+* Technical change.
+* Details:
+  - Fix bootstrap script bug.
```
sandcha: Explicit what this will change for the user?
```suggestion
  - Fix installation and building operations by fixing the bootstrap script.
```

alexsoble:
@sandcha Thank you! Your CHANGELOG suggestion looks good to me. 👍

PR #99, opened 2021-01-29
Added a new variable and associated params for calculating parenting payments

sarafalamaki:
I wanted to have an example to test the new dependencies endpoint in openfisca-core. There wasn't an example that used group entities/group populations there already, so I wrote this one.

Hopefully it'll be useful for people implementing variables that reference more than one entity as well. :-)

I have updated openfisca-core so the tests pass with the new parameters. This is the PR: https://github.com/openfisca/openfisca-core/pull/964

* Tax and benefit system evolution. | Technical improvement. |  Minor change.
* Impacted periods: all.
* Impacted areas: `path/to/file/containing/impacted/variables`
* Details:
  - New feature or new behaviour description
  - Cases for which an error was noticed

- - - -

These changes :

- Impact the OpenFisca-Country-Template public API (for instance renaming or removing a variable)
- Add a new feature (for instance adding a variable)

### DIFF
--- a/openfisca_country_template/variables/benefits.py
+++ b/openfisca_country_template/variables/benefits.py
@@ -69,3 +69,61 @@ class pension(Variable):
         '''
         age_condition = person('age', period) >= parameters(period).general.age_of_retirement
         return age_condition
+
+
+class parenting_allowance(Variable):
+    value_type = float
+    entity = Household
+    definition_period = MONTH
+    label = ''' Allowance for low income people with children to care for.
+                Loosely based on the Australian parenting pension'''
+    reference = \
+        [u"https://www.servicesaustralia.gov.au/individuals/services/centrelink/parenting-payment/who-can-get-it"]
+
+    def formula(household, period, parameters):
+        '''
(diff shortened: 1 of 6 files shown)


Br3nda (approved):
Looks good

openfisca_country_template/variables/benefits.py
```diff
@@ -69,3 +69,45 @@ def formula(person, period, parameters):
...
+        if family_income > income_threshold:
+            return 0
+        is_single = household.nb_persons(Household.PARENT)
+        ages = household.members('age', period)
+        under_8 = household.any(ages < 8)
+        under_6 = household.any(ages < 6)
+        if (is_single and under_8) or under_6:
+            return parameters(period).benefits.parenting_allowance
+        else:
+            return 0
```
sandcha, line 101: Thank you @sarafalamaki for this new feature and sorry because I'm seeing this formula late: every formula should return a numpy array as the households might have different `parenting_allowance` according to their different situations. So there is a bug to fix here 🙂

PR #102, opened 2021-02-24
Adapt formula syntax to vectorial computing

### LINKED ISSUE #100: parenting_allowance returns one value for all households
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:

I read the `parenting_allowance` variable formula.

### Here is what I expected to happen:

I excpected to get a vector with a parenting_allowance for every simulation household as described by [this section](https://openfisca.org/doc/coding-the-legislation/25_vectorial_computing.html#formulas-always-return-vectors) of the documentation.

### Here is what actually happened:

The `return` statement gives one value.

### Here is data (or links to it) that can help you reproduce this issue:

Connected to #99 

## Context

I identify more as a:

- Developer _(I create tools that use the existing OpenFisca code)_.

sandcha:
Fixes #100 

* Bug fix
* Impacted periods: all.
* Impacted areas: `openfisca_country_template/variables/benefits.py`
* Details:
  - Fixes `parenting_allowance` formula syntax
  - Uses vectorial computing to return a `parenting_allowance` amount per household

- - - -

These changes :

- Fix or improve an already existing calculation.

---

Here is an example of YAML test to illustrate two households with different amounts of `parenting_allowance `:

```yaml
- name: h1 household gets parenting_allowance while h2 doesn't as the child is above 8 years old
  period: 2016-12
  input:
    persons: 
      p1:
        age: 42
      p2:
        age: 42
      c1:
        age: 5
      p3:
        age: 42
      c2:
        age: 9
    households:
      h1:
        parents: [ p1, p2 ]
        children: c1
        household_income: 400
      h2:
        parents: p3
        children: c2
        household_income: 300
  output:
    parenting_allowance: [600, 0]
```


openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
+        how much they, and their partner, earn
         if they are single with a child under 8
         or if they are partnered with a child under 6.
         '''
-        family_income = household('household_income', period)
+        household_income = household('household_income', period)
         income_threshold = 500
-        if family_income > income_threshold:
-            return 0
-        is_single = household.nb_persons(Household.PARENT)
```
sandcha, line 94: > This line returned an int value (1 for single, 2 for two parents). As it is always > 0, `is_single` was always True

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
+        income_condition = household_income < income_threshold
+
+        is_single = household.nb_persons(Household.PARENT) == 1
         ages = household.members('age', period)
         under_8 = household.any(ages < 8)
         under_6 = household.any(ages < 6)
-        if (is_single and under_8) or under_6:
-            return parameters(period).benefits.parenting_allowance
-        else:
-            return 0
```
sandcha, line 101: > In both return statements, a unique result was returned. 
But in `def formula(household, period, parameters)` the input `household` is always a numpy array that contains one or multiple households and we want the `parenting_allowance` calculation to take every household specific situation into account => we want a `parenting_allowance` result per household. So, the returned value must be a numpy array.
I tried to give an example in this pull request description with a test on two households with different situations and different `parenting_allowance` results given by one execution of this formula. 😊

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
+
+        allowance_condition = income_condition * ((is_single * under_8) + under_6)
+
+        return allowance_condition * parameters(period).benefits.parenting_allowance
```
sandcha, line 103: Here, the cardinality of the result is given by the `allowance_condition` that has one boolean value per household.
So, even if the parameter is a unique value, thanks to the `*` (aka AND) operation, the result will have a length equal to the number of households.

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
         if they are single with a child under 8
         or if they are partnered with a child under 6.
         '''
-        family_income = household('household_income', period)
+        household_income = household('household_income', period)
         income_threshold = 500
-        if family_income > income_threshold:
-            return 0
-        is_single = household.nb_persons(Household.PARENT)
+        income_condition = household_income < income_threshold
```
sandcha, line 94: > The operator here was `>`. I updated it to `<` as it seemed strange to check for low income family but to give the allowance only if it was **above** an income threshold 🤔

benjello (approved):

openfisca_country_template/variables/benefits.py
maukoquiroga: For the order, I also usually do _real_ vs _expected_, in that order.

For the operator, we're changing behaviour, as it should be _<=_ if we stick to the previous implementation.

Here I understand _no more than_ as either _<=_ or _>_ @sarafalamaki.

Capture d’écran 2021-03-07 à 21 51 52

maukoquiroga: I modified the test to map the _no more than_ behaviour, and it fails as expected.

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
     def formula(household, period, parameters):
         '''
         A person's parenting allowance depends on how many dependents they have,
-        How much they, and their partner, earn
+        how much they, and their partner, earn
         if they are single with a child under 8
         or if they are partnered with a child under 6.
         '''
-        family_income = household('household_income', period)
+        household_income = household('household_income', period)
```
maukoquiroga, line 92: This wont calculate the household's income.

Maybe:

```python
household_income = household.sum(household.members('salary', period))
```

?

openfisca_country_template/variables/benefits.py
maukoquiroga: @sandcha what do you think ?

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
-        family_income = household('household_income', period)
-        income_threshold = 500
-        if family_income > income_threshold:
-            return 0
-        is_single = household.nb_persons(Household.PARENT)
+        parenting_allowance = parameters(period).benefits.parenting_allowance
+
+        household_income = household.sum(household.members('salary', period))
+        income_threshold = parenting_allowance.income_threshold
+        income_condition = household_income < income_threshold
```
maukoquiroga, line 93:
```suggestion
        income_condition = household_income <= income_threshold
```

maukoquiroga (changes requested):
Thanks @sandcha!

I modified the test and the formula a bit to try to reproduce the operator bug I identified, and it worked! Please feel free to drop my commits to solve the problem in another way if you prefer.

Also, while doing that, I identified another bug: we were calling to a non-existing variable _household_income_ without any error or warning!!

I would have expected it to fail...

@sarafalamaki @benjello

openfisca_country_template/variables/benefits.py
```diff
@@ -75,30 +75,32 @@ class parenting_allowance(Variable):
...
         or if they are partnered with a child under 6.
         '''
-        family_income = household('household_income', period)
-        income_threshold = 500
-        if family_income > income_threshold:
-            return 0
-        is_single = household.nb_persons(Household.PARENT)
+        parenting_allowance = parameters(period).benefits.parenting_allowance
+
+        household_income = household.sum(household.members('salary', period))
```
maukoquiroga, line 91: Maybe we should document this ?

openfisca_country_template/variables/benefits.py
sarafalamaki: This looks right to me, and is equivalent to the original, where we exited with 0 allowance if the income was greater than the threshold. This is a bit clearer though, and using the parameter is definitely better than my lazy approach.

openfisca_country_template/variables/benefits.py
```diff
@@ -105,31 +105,30 @@ def formula(household, period, parameters):
...
-        is_single = household.nb_persons(Household.PARENT)
+        is_single = household.nb_persons(Household.PARENT) == 1
         ages = household.members("age", period)
         under_8 = household.any(ages < 8)
         under_6 = household.any(ages < 6)
 
-        if (is_single and under_8) or under_6:
-            return parameters(period).benefits.parenting_allowance
+        allowance_condition = income_condition * ((is_single * under_8) + under_6)
+        allowance_amount = parenting_allowance.amount
```
sarafalamaki, line 120: Why do you recommend using arithmetic operators rather than logical ones here?

In my experience, this lead to hard to see bugs when done in more complex situations.

openfisca_country_template/variables/benefits.py
sarafalamaki: Absolutely, clearly a braino on my part!

sarafalamaki (approved):

maukoquiroga (approved):

PR #113, opened 2021-08-05
Web API smoke test

### LINKED ISSUE #112: Smoke test the Web API
There was a production incident last week on OpenFisca France, where an inadequate config key prevented the Web API from starting altogether. This would have been caught if the CI had a small test for the web API.

I thus suggest to add a smoke test to the web API, which will check:

1. The API starts with `openfisca serve`.
2. Accessing the `swagger` endpoint returns an HTTP 200 code.
3. The result of the `swagger` endpoint is valid JSON.

This test result should appear as a dedicated step in each PR, along with `check_version_and_changelog` or `lint_yaml_files`.

HAEKADI:
Closes #112.
 
* Technical improvement.
* Details:
  - Add a Web API smoke test to the CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,9 +62,13 @@ jobs:
           key: v1-py3-deps-{{ checksum "setup.py" }}
 
       - run:
+          name: Avtivate virtualenv
+          command: |
+            echo "source /tmp/venv/country_template/bin/activate" >> $BASH_ENV
+
+      - run:
           name: Test the Web API
           command: |
-            source /tmp/venv/country_template/bin/activate
             .circleci/test-api.sh
 
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,16 @@ jobs:
             git fetch
             .circleci/is-version-number-acceptable.sh
 
+  test_api:
+    docker:
+      - image: python:3.7
+    steps:
+      - checkout
+
+      - run:
+          name: Test the Web API
+          command: .circleci/test-api.sh
+
   deploy:
     docker:
       - image: python:3.7
@@ -86,6 +96,9 @@ workflows:
   build_and_deploy:
     jobs:
       - build
+      - test_api:
+          requires:
+            - build
       - check_version
       - deploy:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,15 +58,14 @@ jobs:
     steps:
       - checkout
 
-      - run:
-         name: Install dependencies
-         command: |
-            make build
-            # pip install --editable git+https://github.com/openfisca/openfisca-core.git@BRANCH#egg=OpenFisca-Core[web-api]  # use a specific branch of OpenFisca-Core
+      - restore_cache:
+          key: v1-py3-deps-{{ checksum "setup.py" }}
 
       - run:
           name: Test the Web API
-          command: .circleci/test-api.sh
+          command: |
+            source /tmp/venv/country_template/bin/activate
+            .circleci/test-api.sh
 
   deploy:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,13 +63,11 @@ jobs:
 
       - run:
           name: Avtivate virtualenv
-          command: |
-            echo "source /tmp/venv/country_template/bin/activate" >> $BASH_ENV
+          command: echo "source /tmp/venv/country_template/bin/activate" >> $BASH_ENV
 
       - run:
           name: Test the Web API
-          command: |
-            .circleci/test-api.sh
+          command: .circleci/test-api.sh
 
   deploy:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
           key: v1-py3-deps-{{ checksum "setup.py" }}
 
       - run:
-          name: Avtivate virtualenv
+          name: Activate virtualenv
           command: echo "source /tmp/venv/country_template/bin/activate" >> $BASH_ENV
 
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,12 @@ jobs:
       - checkout
 
       - run:
+         name: Install dependencies
+         command: |
+            make build
+            # pip install --editable git+https://github.com/openfisca/openfisca-core.git@BRANCH#egg=OpenFisca-Core[web-api]  # use a specific branch of OpenFisca-Core
+
+      - run:
           name: Test the Web API
           command: .circleci/test-api.sh
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,12 +54,10 @@ jobs:
   test_api:
     docker:
       - image: python:3.7
+
     steps:
       - checkout
-
-      - restore_cache:
-          key: v1-py3-deps-{{ checksum "setup.py" }}
-
+      
       - run:
           name: Test the Web API
           command: .circleci/test-api.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,9 @@ jobs:
     steps:
       - checkout
 
+      - restore_cache:
+          key: v1-py3-deps-{{ checksum "setup.py" }}
+
       - run:
           name: Test the Web API
           command: .circleci/test-api.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
 
     steps:
       - checkout
-      
+
       - run:
           name: Test the Web API
           command: .circleci/test-api.sh
--- a/.circleci/test-api.sh
+++ b/.circleci/test-api.sh
@@ -6,7 +6,7 @@ apt-get install -y jq
 make serve-local &
 
 PORT=5000
-ENDPOINT=specs # wrong, to test the --fail option
+ENDPOINT=spec
 
 response=$(curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | jq -e)
 
--- a/.circleci/test-api.sh
+++ b/.circleci/test-api.sh
@@ -4,5 +4,11 @@ PORT=5000
 ENDPOINT=spec
 
 openfisca serve --country-package openfisca_country_template --port $PORT &
+server_pid=$!
 
 curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | python -m json.tool > /dev/null
+result=$?
+
+kill $server_pid
+
+exit $?
--- a/.circleci/test-api.sh
+++ b/.circleci/test-api.sh
@@ -1,8 +1,8 @@
 #! /usr/bin/env bash
 
-make serve-local &
-
 PORT=5000
 ENDPOINT=spec
+
+openfisca serve --country-package openfisca_country_template --port $PORT &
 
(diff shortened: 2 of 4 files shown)


.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env bash
+
+apt-get update
+apt-get install -y jq
+
+make serve-local &
+
+curl --retry-connrefused \
+    --retry 5 \
+    --retry-delay 20 \
```
MattiSG, line 10: 20 seconds of delay seems long, especially if we try 5 times: if the result is negative, we will systematically have a delay of 20×5 = 100 seconds = 1 minute 40 seconds.

Choose instead the median delay for the API start-up  :)
```suggestion
    --retry-delay 5 \
```

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+    http://127.0.0.1:5000 --fail
+
+var=$?
```
MattiSG, line 13: This varname is not explicit.
```suggestion
result=$?
```

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+
+apt-get update
+apt-get install -y jq
+
+make serve-local &
+
+curl --retry-connrefused \
+    --retry 5 \
+    --retry-delay 20 \
+    http://127.0.0.1:5000 --fail
```
MattiSG, line 11: Consistency: either every option should be on its own line, or they should be on one line 🙂

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env bash
+
+apt-get update
+apt-get install -y jq
+
+make serve-local &
```
MattiSG, line 6: For test resilience, do not rely on default port and host: define at least `PORT` and use it both as a parameter to `serve-local` and in the tested `curl` target.

MattiSG: Why not `openfisca serve`? 🙂

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+    --retry-delay 20 \
+    http://127.0.0.1:5000 --fail
+
+var=$?
+
+if [[ $var -ne 0 ]]
+then
+  echo "Server failed to start."
+  exit $var
+fi
```
MattiSG, line 19: > All of this could be more concisely expressed as `curl …… || echo 'Web API does not reply'` 🙂

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+curl --retry-connrefused \
+    --retry 5 \
+    --retry-delay 20 \
+    http://127.0.0.1:5000 --fail
+
+var=$?
+
+if [[ $var -ne 0 ]]
+then
+  echo "Server failed to start."
```
MattiSG, line 17: Is it really that the server failed to start, or that it is unreachable? 🙂 We don't know if it did start or not.

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+  exit $var
+fi
+
+response=$(curl --write-out '%{http_code}' --silent --output /dev/null http://127.0.0.1:5000/spec)
```
MattiSG, line 21: Performance: why make two requests instead of using this request to check for availability? 🙂

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
...
+if [[ $response -ne 200 ]]
+then
+  echo "Accessing the swagger endpoint returns an HTTP $response code instead of a 200 code."
+  exit 1
+fi
```
MattiSG, line 26: I understand this is already what the `--fail` option does, as can be seen with `curl https://httpstat.us/400 --fail` 🙂

.circleci/test-api.sh
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env bash
+
+apt-get update
+apt-get install -y jq
```
MattiSG, line 4: Installing dependencies should not be done on every test run, only in the dependencies step of the CI config 🙂
If this test is run on a local machine, for example, it might not have access to `apt-get`, or have `jq` already installed.

.circleci/test-api.sh
MattiSG: If we store this in a separate shell script, we support running the tests outside from CI. I think it is great, but in this case we have to kill this process after it was created, otherwise the host machine will keep the process running.

MattiSG (changes requested):
Nice! I think this is functional, but has a lot of complexity. I'm pretty sure all tests can end up being pretty much a one-liner, perhaps along the lines of:

```bash
PORT=5000
ENDPOINT=swagger

set -e
curl --retry-connrefused --retry 5 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | jq -e
```

.circleci/test-api.sh
HAEKADI: For some reason that I don't understand, `openfisca serve` does not work. Only `make serve-local` does, even though all it does is call `openfisca serve`. I tried restoring the cache from the `build` job in the `test_api` but it did not help.
Here are the [logs](https://app.circleci.com/pipelines/github/openfisca/country-template/105/workflows/2d3555e4-8e90-4998-af90-841596e7e44e/jobs/1086) `.circleci/test-api.sh: line 6: openfisca: command not found`. 
The job still succeeded but that is another problem that I will discuss below.

.circleci/test-api.sh
HAEKADI: I updated it to --retry 10, because 5 is not enough for the server to start up.

.circleci/test-api.sh
HAEKADI: The idea as we discussed was to have explicit error messages :)

.circleci/test-api.sh
HAEKADI: From what I understand, the `--fail` option fails silently (no output at all) [on server errors](https://curl.se/docs/manpage.html), hence the following [run](https://app.circleci.com/pipelines/github/openfisca/country-template/107/workflows/5f4922e6-f97f-4314-a59b-d203f12a9dba/jobs/1092), where I purposely gave the [wrong endpoint](https://github.com/openfisca/country-template/commit/71b2c3a0ad1d825bceecc6613b2c721d160b52af#diff-74fe81df0a9ccb18afb1539db65408e43bb8924e6dbd235ddf00073e6a4200f2R9) and the test still succeeded in the CI even after having a ```curl: (22) The requested URL returned error: 404 NOT FOUND```.
[--fail-with-body ](https://curl.se/docs/manpage.html) seemed like an option to solve the problem, however it's only available for `curl 7. 76+ and I have `curl 7.64.1`.
I then [assigned the curl command to a variable](https://github.com/openfisca/country-template/commit/26712d9e371c5824a94ab244a1789f218c52f936) and that seems to solve the issue. I now get an [error](https://app.circleci.com/pipelines/github/openfisca/country-template/108/workflows/4201355b-4fbb-4b1d-b7f0-1b3b6bd1c27b/jobs/1095).
What do you think ? :)

HAEKADI:
@MattiSG I have an issue with the last commit. I could use some help :)
When I ran the following `test_api` script:
```
#! /usr/bin/env bash

apt-get update
apt-get install -y jq

make serve-local &

PORT=5000
ENDPOINT=spec

set -e 
response=$(curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | jq -e)
```
I got an error:
```
curl: (23) Failed writing body (0 != 15248)
```
I tried : 
```
response=$(curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | tac | tac | jq -e)
```
And I got the same error.

The following command runs with no issues:
```
curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | jq -e
```
Do you have any idea why that is?

.circleci/test-api.sh
MattiSG: I don't understand why the variable assignment would be enough to change the exit code 🤔

> fails silently (no output at all)

To _fail_ means having an exit code > 0, which is different from “silently”, with no output.
In this case, it is not even _silent_ as it outputs the HTTP status code and status name. The `--fail` option only stops the server page from being shown, and changes the exit code, which is what we want.

In the current case, I guess `jq` is happy with an empty input and will not fail. We thus have to either `set -o pipefail` or just avoid the pipe with an intermediate variable:

```bash
set -e

response=$(curl ……)
jq $response
```

[More information](https://unix.stackexchange.com/a/73180).

HAEKADI:
`jq` remains problematic even after adding an intermediary variable; rendering the following error, which I could not fix:
```
jq: error: syntax error, unexpected $end, expecting QQSTRING_TEXT or QQSTRING_INTERP_START or QQSTRING_END (Unix shell quoting issues?) at <top-level>
``` 
So I switched to `python -m json.tool` which works, and as a bonus there is no need to add a `jq`dependency.

.circleci/test-api.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env bash
+
+make serve-local &
+
+PORT=5000
+ENDPOINT=spec
+
+curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | python -m json.tool
```
MattiSG, line 8: > `python -m json.tool` pretty-prints the resulting JSON. I'm afraid that might be a bit of a large log on larger tax and benefit systems. Maybe we should silence the standard output?

.circleci/test-api.sh
MattiSG: The two other comments above are still valid 🙂

MattiSG (changes requested):
Getting there!  🙂

.circleci/config.yml
```diff
@@ -51,6 +51,24 @@ jobs:
...
+      - image: python:3.7
+
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-py3-deps-{{ checksum "setup.py" }}
+
+      - run:
+          name: Avtivate virtualenv
```
MattiSG, line 65:
```suggestion
          name: Activate virtualenv
```

.circleci/test-api.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env bash
+
+PORT=5000
+ENDPOINT=spec
+
+openfisca serve --country-package openfisca_country_template --port $PORT &
+
+curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | python -m json.tool > /dev/nul
```
MattiSG, line 8:
```suggestion
curl --retry-connrefused --retry 10 --retry-delay 5 --fail http://127.0.0.1:$PORT/$ENDPOINT | python -m json.tool > /dev/null
```

MattiSG: This does work because `curl --fail` will not forward the (potentially valid JSON) response, as evidenced by:

```
› curl --header 'Accept:application/json' --retry-connrefused --retry 1 --retry-delay 1 --fail https://httpstat.us/418 | python -m json.tool
curl: (22) The requested URL returned error: 418 
No JSON object could be decoded
```

MattiSG (changes requested):
I'll handle the minor corrections.

MattiSG (approved):
🎉 🚀

HAEKADI:
@MattiSG Il faudera rajouter `test_api` comme required dans les status checks avant le merge, non ?

MattiSG:
Good point! Done now :)

PR #111, opened 2021-05-03
Improve packaging

bonjourmauko:
#### Patch changes

- Improve `pytest` support for `doctest`
- Use `pip check` for dependency conflicts
- Add `pyupgrade` for automatic code formatting
- Add `darglint` for improved docstring linting
- Add `.editorconfig` for shared editor configuration
- Add `pre-commit` for automatic style checks & formatting
  - Git hooks are optional, as they may not suit everybody
  - To enable them, run `pre-commit install`
- Add long description to package distribution
- Update distribution keywords

#### Minor changes

- Relax upper-bound support for OpenFisca-Core
  - This package and OpenFisca-Core are mutually dependent, which breaks `pip`
  - The least astonishing is to assume this package to always be compatible with core
  - This is also easily enforceable thanks to regular dependabot checks that will run in CircleCI
  - Finally, it is the responsibility of the OpenFisca-Core integrity checks workflow to ensure this package to always be compatible
    - Every new version of core has to ensure all of the tests included in and distributed by this package pass
    - Also, every breaking change to OpenFisca-Core has to be phased out to help adapting this package to those changes

#### Major changes

- Drop support for Python < 3.7

PR #115, opened 2021-09-02
Switch from CircleCI to GitHub Actions

### LINKED ISSUE #114: Lint YAML and Python files on CI
OpenFisca-France has tests to automatically lint [YAML](https://github.com/openfisca/openfisca-france/blob/master/.circleci/lint-changed-yaml-tests.sh) and [Python](https://github.com/openfisca/openfisca-france/blob/master/.circleci/lint-changed-python-files.sh) files.

I believe the publishing process is already part of the OpenFisca Country Template suggestions, and that maintaining consistent quality and syntax is a meaningful way to ease contribution.

I thus suggest importing these tests, alongside any other improvements made to the testing procedure, to the Country Template (probably toning down a bit the tongue-in-cheek phrasings 😉).

Opinions welcome @openfisca/international-contrib!

HAEKADI:
Related to [openfisca/openfisca-core/#1030](https://github.com/openfisca/openfisca-core/issues/1030)
Related to [openfisca/openfisca-core/#1027](https://github.com/openfisca/openfisca-core/issues/1027)
Related to [openfisca/openfisca-france/#1663](https://github.com/openfisca/openfisca-france/pull/1663)
Closes #114 

[GitHub Actions](https://github.com/openfisca/country-template/actions)

* Technical improvement. 
* Impacted areas: CI configuration
* Details:
  - Switch from CircleCI to GitHub Actions
 
Screenshot 2021-09-23 at 15 47 48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.13.0 - [#115](https://github.com/openfisca/country-template/pull/115)
+
+* Technical improvement.
+* Details:
+  - Switch continuous integration pipeline from CircleCI to GitHub Actions 
+
 ### 3.12.10 - [#119](https://github.com/openfisca/country-template/pull/119)
 
 * Technical improvement.


HAEKADI:
@MattiSG @maukoquiroga `check-style` in the `Makefile` uses both `pylint` and `flake8` in [`Country-template`](https://github.com/openfisca/country-template/blob/62a183ddd1da6580f839f3b70fab7259270f0d92/Makefile#L34-L38) but only `flake8` in [`OpenFisca France`](https://github.com/openfisca/openfisca-france/blob/03d859105fa245ae09055e190ea9420e864681c8/Makefile#L36-L39). 

The old CI did not include a linting job #114, when adding it to the new CI `pylint` causes the following error:

Screenshot 2021-09-15 at 11 02 05

Is there a consensus on whether one or both should be used?

MattiSG:
Unless they do different things, I am in favour of having only one linter, to make linting faster and more deterministic.
I see that on Core, there's only `flake8`.

@HAEKADI didn't you mention `superlinter`? Do you think it would be appropriate here?

HAEKADI:
@MattiSG Yes, I have. It would definitely be appropriate. You can find the tests I ran with it [here](https://github.com/openfisca/country-template/runs/3493407898?check_suite_focus=true). 

But if my memory serves me well, we decided to add it in a separate PR after the switch to GitHub Actions as to not introduce too many changes at once.

MattiSG:
> we decided to add it in a separate PR after the switch to GitHub Actions as to not introduce too many changes at once

We did indeed 🙂 Now, if we have to reconsider the current way of linting, I just wanted to throw this in 😉

maukoquiroga:
Hi !

> Is there a consensus on whether one or both should be used?

We have agreed on the rules to encforce, that's why you will find them in the country and extension templates, yet still not implemented them in openfisca/openfisca-core#960 . 

> Unless they do different things, I am in favour of having only one linter, to make linting faster and more deterministic.
> I see that on Core, there's only `flake8`.

`flake8` and `pylint` are lint "engines", they implement style codes, like PEP8 and so on.

What they do differently is hang indents (`E128/133`), a very specific style of the older OpenFisca community.

Other than that, I prefer whatever comes closes to the rules we have agreed upon (that excludes `black` for example).

Specific example: in OpenFisca, a `Variable` is a class, but it is snake cased, so the only prerequisite of a linter for me is that it can be highly configurable.

> @HAEKADI didn't you mention `superlinter`? Do you think it would be appropriate here?

Haven't heard of that one, I'll take a look.

The minimal set of style rules for me is the one in the setup.cfg, well the exceptions:

```
; D101:                   Variables already provide label/description
; D107:                   We do not document __init__ method
; D401:                   We do not require the imperative mood
; E128/133:               We prefer hang-closing visual indents
; E251:                   We prefer `function(x = 1)` over `function(x=1)`
; E501:                   We do not enforce a maximum line length
; W503/4:                 We break lines before binary operators (Knuth's style)

; C0103:                  We (still) snake case variables and reforms
; C0115:                  Variables already provide label/description
; C0301:                  We do not enforce a maximum line length
; E0213:                  This requires changes in OpenFisca-Core
; E1101:                  False positive, as entities have members
; E1102:                  False positive, as entities are callable
; W0621:                  We name variable values the variable name
; W1203:                  We prefer to log with f-strings
```

HAEKADI:
Thank you for the clarification @maukoquiroga! I started implementing `super-linter` in `Openfisca France` [here](https://github.com/openfisca/openfisca-france/pull/1679).
I used super-linter not only for new scripts but on the existing code-base and there were too many [errors](https://github.com/openfisca/openfisca-france/runs/3630266106?check_suite_focus=true) to count.
I was not aware that `black` was excluded and consequently started correcting those errors 😬

maukoquiroga:
I don't have a clear view on latest discussions in the French community, but for the historical record `black` has been rejected twice while I was leading product. I was agnostic since, now I'm more on the reluctant side.

> used super-linter not only for new scripts but on the existing code-base and there were too many [errors](https://github.com/openfisca/openfisca-france/runs/3630266106?check_suite_focus=true) to count.

I can relate to that, see https://github.com/openfisca/openfisca-core/pull/960 and https://github.com/openfisca/openfisca-france/issues?q=style+author%3Amaukoquiroga.

HAEKADI:
I see. Maybe it's time to put pen to paper and have a clear definition of linting rules across all repos. I see you already started working on it @maukoquiroga with [openfisca/openfisca-core#960](https://github.com/openfisca/openfisca-core/pull/960).

maukoquiroga:
> I see. Maybe it's time to put pen to paper and have a clear definition of linting rules across all repos. I see you already started working on it @maukoquiroga with [openfisca/openfisca-core#960](https://github.com/openfisca/openfisca-core/pull/960).

That would be great!!!

See one of the discussions on `black` for example: https://github.com/openfisca/openfisca-core/pull/977

I've started https://github.com/openfisca/openfisca-core/blob/master/STYLEGUIDE.md and I'm happy to help if you decide to continuing efforts on style coherence —which I think is a great idea.

It is an idea that has been going on and proposed by some members. See https://github.com/openfisca/country-template/pull/111/files for example, beyond the actual implementation (`pre-commit`), it would be great to have all linting rules in a separated repo and pull them as a development dependency.

HAEKADI:
I see. There are certainly many ways to go about doing that.
As a starting point, I suggest implementing [these](https://github.com/openfisca/country-template/blob/master/setup.cfg) rules, in the new CI, with [super-linter](https://github.com/openfisca/openfisca-france/blob/abdaa6110e05221114eeda3b107795978a6680ee/.github/workflows/workflow.yml#L31-L48),  across all repos ([France](https://github.com/openfisca/openfisca-france/pull/1679), Country and Core) as discussed before, to have a unified style.

MattiSG:
If I understand correctly, [`super-linter`](https://github.com/github/super-linter) is just a way to ease working with linters on GitHub Actions:

> The design of the Super-Linter is currently to allow linting to occur in GitHub Actions as a part of continuous integration occurring on pull requests as the commits get pushed. It works best when commits are being pushed early and often to a branch with an open or draft pull request. There is some desire to move this closer to local development for faster feedback on linting errors but this is not yet supported.

However, in https://github.com/openfisca/openfisca-core/issues/1040, we decided to focus on consolidating tasks in a task manager, not in CI. Thus, I understand that if we want to call both `flake8` and `pylint`, this should be wrapped in a task that is also executable locally and whose dependencies are declared and installed explicitly, not only indirectly through `super-linter`.

I also see in https://github.com/openfisca/country-template/pull/115/commits/aaea4ddd6e0566609b5a1265d40f07a6da6fc4f4 that the so-called “linting scripts” were removed, but these scripts do more than just activate linting: most of all, they select the subset of files to be linted in order to accelerate the process. I am thus not sure this should just be removed.

MattiSG:
As discussed this morning with @HAEKADI, based on the conclusions from https://github.com/openfisca/openfisca-core/issues/1040, let's focus this PR on being isofunctional and moving to GitHub Actions without changing the services.

HAEKADI:
I removed super-linter. 
I just have to check `deploy`. @MattiSG could you please add the pypi password in GitHub secrets?

MattiSG:
> could you please add the pypi password to GitHub secrets?

Done!

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,155 @@
...
+        with:
+          python-version: 3.7
+      - name: Cache build
+        id: restore-build
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ github.ref }}${{ env.pythonLocation }}-build-${{ hashFiles('setup.py') }}
+      - run: |
+          openfisca test --country-package openfisca_country_template openfisca_country_template/tests
```
MattiSG, line 71:
```suggestion
      - run: openfisca test --country-package openfisca_country_template openfisca_country_template/tests
```

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,155 @@
...
+          python-version: 3.7
+      - name: Check version number has been properly updated
+        run: "${GITHUB_WORKSPACE}/.github/is-version-number-acceptable.sh"
+
+  # GitHub Actions does not have a halt job option, to stop from deploying if no functional changes were found.
+  # We build a separate job to substitute the halt option.
+  # The `deploy` job is dependent on the output of the `check-for-functional-changes`job.
+  check-for-functional-changes:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/github-actions' # Only triggered for the `master` branch
```
MattiSG, line 110: Reminder to update 😛

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,155 @@
...
+          path: ${{ env.pythonLocation }}
+          key: ${{ github.ref }}${{ env.pythonLocation }}-build-${{ hashFiles('setup.py') }}
+      - name: Cache build release
+        id: restore-build-release
+        uses: actions/cache@v2
+        with:
+          path: dist
+          key: ${{ env.pythonLocation }}-build-release-${{ hashFiles('setup.py') }}
+      - name: Upload a Python package to PyPi
+        run: twine upload dist/* --username "$PYPI_USERNAME" --password "$PYPI_PASSWORD"
```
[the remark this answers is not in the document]
HAEKADI: It's installed in the `build` with the `deps` step :)

.github/is-version-number-acceptable.sh
```diff
@@ -12,12 +12,12 @@ then
...
 
-current_version=`python setup.py --version`
+current_version=$(python setup.py --version)
 
-if git rev-parse --verify --quiet $current_version
+if git rev-parse --verify --quiet "$current_version"
 then
     echo "Version $current_version already exists in commit:"
-    git --no-pager log -1 $current_version
+    git --no-pager log -1 "$current_version"
```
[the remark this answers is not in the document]
HAEKADI: Actually, it happened when I tested super-linter; I got many `shellcheck reported issue in this script: SC2086:info:1:1: Double quote to prevent globbing and word splitting [shellcheck]` errors, hence the quotes :)

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,155 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Fetch all the tags
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - id: stop-early
+        run: if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh" ; then echo "::set-output name=status::success" ; fi
```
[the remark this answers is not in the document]
HAEKADI: Yes, indeed. Nice catch :)

HAEKADI:
The `deploy` seems functional.
Screenshot 2021-09-23 at 17 19 56

setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name = "OpenFisca-Country-Template",
-    version = "3.12.9",
+    version = "3.12.10",
```
MattiSG, line 7: In the context of the Template, I believe that switching from CircleCI to GitHub Actions should be a minor, not a patch 🙂

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+### 3.12.10 - [#115](https://github.com/openfisca/country-template/pull/115)
+
+* Technical improvement.
+* Details:
+  - Switch CI from CircleCI to GitHub Actions
```
MattiSG, line 7:
```suggestion
  - Switch CI provider from CircleCI to GitHub Actions 
```

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,153 @@
...
+          python-version: 3.7.12
+      - name: Cache build
+        id: restore-build
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ github.ref }}${{ env.pythonLocation }}-build-${{ hashFiles('setup.py') }} # Cache the entire build Python environment
+      - name: Build package
+        if: steps.restore-build.outputs.cache-hit != 'true'
+        run: make build
```
MattiSG, line 23: Won't that leave the built package in its very first state, never updating the cache? 🤔

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,153 @@
...
+    runs-on: ubuntu-latest
+    needs: [ build ]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Fetch all the tags
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7.12
```
MattiSG, line 41: Do we really need to specify down to the patch version? I would prefer if we could simply define `3.7` and work on the latest patch version 🙂

MattiSG (changes requested):
The workflow name is not updated when the bootstrap script is executed: the `workflow.yml` file still contains `name: Country Template`.

.github/workflows/workflow.yml
HAEKADI: Yes. Because the name of the environment in the cache depends on the patch as well. And in `Openfisca-France' there were issues where the build used 3.7.12 and the test used 3.7.11 and it couldn't find the right cache 🤔

setup.py
HAEKADI: Okay :) so `3.13.0` ?

.github/workflows/workflow.yml
HAEKADI: The cache is updated if any element of the cache key is changed, in this case the `setup`, the `branch` or the `python env`. Do you think something else should be added?

HAEKADI: Yes, you're right. We did encounter the problem in `Openfisca-France`. I opened a [PR](https://github.com/openfisca/openfisca-france/pull/1685) to solve it.

HAEKADI: I will integrate the changes here as well.

bootstrap.sh
```diff
@@ -36,8 +36,9 @@ all_module_files=`find openfisca_country_template -type f`
 set -x
 
 # Use intermediate backup files (`-i`) with a weird syntax due to lack of portable 'no backup' option. See https://stackoverflow.com/q/5694228/594053.
-sed -i.template "s|country_template|$lowercase_country_name|g" README.md setup.py .circleci/config.yml Makefile MANIFEST.in $all_module_files
-sed -i.template "s|Country-Template|$COUNTRY_NAME|g" README.md setup.py .github/PULL_REQUEST_TEMPLATE.md CONTRIBUTING.md $all_module_files
+# If you're using MacOS, add LC_ALL=C before sed https://stackoverflow.com/questions/19242275/re-error-illegal-byte-sequence-on-mac-os-x/19770395#19770395
```
MattiSG, line 39: I don't have any warning or issue with macOS 11.6. What version did trigger the addition of this comment? 🙂
In any case, the aim with this bootstrapping script is to be fully automatic. If we need adjustments, they should be automated, we cannot expect users to edit the script itself.

MattiSG (approved):
Tested locally entirely on macOS 11.6 with an M1 processor, the bootstrap script seems to work well, but installing fails with `ModuleNotFoundError: No module named 'Cython'` when trying to install `numexpr 2.7.3`.

Since the tests pass, I assume this is related to my local setup and will approve this PR. I would appreciate a double check on this issue, as well as handling the macOS-specific comment 🙂

Looking forward to this landing!

PR #121, opened 2021-10-26
Update deprecated API

maukoquiroga:
Depended upon by openfisca/openfisca-core#1015

* Technical improvement.
* Details:
  - Update code using deprecated parts of OpenFisca-Core Public API

- - - -

These changes:

- Adapt to the current OpenFisca-Core API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 3.13.2 - [#121](https://github.com/openfisca/country-template/pull/121)
+
+* Minor change
+* Details:
+  - Delete macOS related comment in bootstrap script
+
 ### 3.13.1 - [#120](https://github.com/openfisca/country-template/pull/120)
 
 * Minor change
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+### 3.13.2 - [#121](https://github.com/openfisca/country-template/pull/121)
+
+* Minor change
+* Details:
+  - Delete macOS related comment in bootstrap script
```
HAEKADI, line 7: There was a mix-up with the previous minor version I think 😅

HAEKADI (approved):
I ran the tests and it looks good to me 🚀

CHANGELOG.md
MattiSG: This description is not correct 😉

MattiSG (changes requested):
Since this update is made necessary to reflect an OpenFisca-Core version change, the OpenFisca-Core dependency version should be bumped 🙂

MattiSG:
Even though the tests pass, I don't understand how this changeset can be compatible with the current version of Core since only in https://github.com/openfisca/openfisca-core/pull/1015/files#diff-f81fa9f88e2aa6671067dccf0760afe4aca86ef0e6f4a03ee9da39a061077043 is the `model_api` importing `ParameterScaleBracket` 🤔

maukoquiroga:
Superseded by #127

PR #125, opened 2022-09-02
Test core and survey-manager evolution

benjello:
Build on a specific version of core

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,8 @@ setup(
             ),
         ],
     install_requires = [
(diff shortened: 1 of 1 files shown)


MattiSG:
Hi @benjello, this PR seems to have very limited reuse value considering its name and current changeset. I will close it without deleting the branch, so you can still work on your experimentations while keeping the number of open PRs low 😊
Feel free to reopen if it is important to you that this stays open as a draft!

PR #130, opened 2023-04-27
Allows country-template to pass check-style makefile command

verbman:
Allows country-template to pass check-style makefile command out of the box

* Technical improvement.
* Impacted periods: none.
* Impacted areas: all
* Details:
  - If you follow the make new country template instructions and at the end run `make test` there are style errors. This resolves those errors for that satisfying green tick.

- - - -

These changes:

- Fix or improve an already existing calculation.
- Change non-functional parts of this repository (for instance editing the README)

PR #132, opened 2023-06-03
Ensure newly generated country package passes check-style makefile command

### LINKED ISSUE #131: Linting issue in new country template
Hi there!

I really enjoy OpenFisca, but I recently encountered an issue.

### Here is what I did:
Created a new country repository from the OpenFisca country template

### Here is what I expected to happen:
The linting tests to pass when I ran `make test` locally

### Here is what actually happened:
The linting tests failed

### Here is data (or links to it) that can help you reproduce this issue:
I believe the issue originates in the `setup.cfg` file as it still references `openfisca_country_template` on lines 17,18 and 36 in my new country repository created utilising the `bootstrap.sh` script from the openfisca_country_template repository.

The way to resolve this issue will be to edit the `bootstrap.sh` [to include `setup.cfg` - perhaps on line 39](https://github.com/openfisca/country-template/blob/master/bootstrap.sh#LL39C1-L39C152) as per the following untested suggestion:

```
sed -i.template "s|country_template|$lowercase_country_name|g" README.md setup.py setup.cfg .github/workflows/workflow.yml Makefile MANIFEST.in $all_module_files
```

## Context

I identify more as a:
- Developer

Harshil-Jani:
Fixing the linting errors based on suggestions give by @verbman in #130 
Fixes #131 

- - - -

- Correct the bootstrap script


MattiSG:
Thanks @Harshil-Jani! 🙏

@verbman since you raised the issue, would you have time to check if that does fix #131? 🙂

MattiSG:
@Harshil-Jani I see that you also fixed a typo in this changeset. Thank you very much for spotting it and correcting it! 🙏

While I understand this has a higher administrative cost, in the future, please separate concerns in each changeset 🙂 This eases review for maintainers, as they can focus on only one topic; and it increases the likelihood of integrating your pull requests and the pace at which we can do so, as some parts might need a lighter review. In the case at hand, for example, we could already merge your typo correction, but it is blocked because of the need for a longer review of the bootstrap script 🙂

MattiSG (commented):
This seems good, thanks @Harshil-Jani! Did you test generating from the template locally after this change? 🙂 

We will need an entry in the CHANGELOG before approval. This change should be a patch change.

Harshil-Jani:
> Did you test generating from the template locally after this change? slightly_smiling_face

Ahm, Nope. I am really going short of memory on my partition and had some weekend work so haven't got time to clean it up and thus I ran out of space when running a virtual env. I was thinking to leverage the github CI (Free hardware 😀 ) to test this PR although the change was regarding local make command but, I thought, If CI is good to go, I can in upcomming week test for local by doing some nits over my disk.

> in the future, please separate concerns in each changeset slightly_smiling_face This eases review for maintainers, as they can focus on only one topic;

Yes agreed. I just reverted my change for the spelling error and will do that in different PR or if you suggest in this PR itself with a different commit whichever sounds best.

MattiSG:
> I was thinking to leverage the github CI

Unfortunately the CI workflow does not go all the way to executing the template generator and testing the resulting new country model. This would be a cool idea though!

Thank you for having updated your commits 🙂 

Someone will try and execute the template locally and review your contribution.

MattiSG:
I reviewed this locally and could confirm that the replacement are effective and no obvious negative side effects is visible.
Unfortunately, I could not go all the way to test that the post-replacement API works properly, as I got blocked by `  ERROR: Failed building wheel for numpy`.

I assume this could be fixed by #129.

I suggest to first fix the linting issue identified by the CI and rebase this PR on top of `master`, and then to try again 🙂

Harshil-Jani:
> I suggest to first fix the linting issue identified by the CI and rebase this PR on top of `master`, and then to try again slightly_smiling_face

Thankyou for the review and helping out. 
I think linting issue was already addressed in #129 and just rebasing on master would be enough.

MattiSG:
Indeed, thanks @Harshil-Jani! The last blocking point seems to be the changelog and version bump, for which you can get instructions [here](https://github.com/openfisca/country-template/blob/master/CONTRIBUTING.md#advertising-changes).
I am testing locally with the updates and will report as soon as this is done 🙂

MattiSG (approved):
Wonderful, after checking locally, this changeset does enable passing `make check-style` after initialisation!
Beyond that, it fixes an important missing variable replacement.
Thanks @Harshil-Jani!

MattiSG:
As the CI says, you just need to reflect the version number bump in `setup.py` for this to be merged 🙂

PR #143, opened 2024-04-26
Automate template setup process via CI for GitHub users

Ndpnt:
* Technical improvement
* Minor change
* Impacted areas: Setup script and CI workflows
* Details:
  - Automate template setup process via CI for GitHub users:
    - When creating a new repository by using this template on GitHub, the setup script is automatically executed by the CI on the resulting generated repository.
  - Improve accessibility by adopting less technical terminology:
    - Replaced "bootstrap" with "first-time setup," clarifying its purpose as a one-time project initialization, particularly useful for users unfamiliar with the technical vocabulary.
  - Decompose GitHub Actions monolithic workflow into specialized workflows:
    - Split the `Country-Template` workflow into three distinct workflows: `build`, `validate` and `deploy` enhancing clarity and organization.
  - Enhance accuracy of workflow triggers:
    - Trigger deployment exclusively when a PR is merged on the `main` branch
    - Trigger validation on PR events or as a dependency of the deployment workflow
    
- - - -

These changes:

- Change non-functional parts of this repository: CI, setup script and documentation

The automation of template setup process via CI can be tested on my fork [here](https://github.com/new?template_name=country-template&template_owner=Ndpnt)


README.md
```diff
@@ -1,25 +1,42 @@
...
 
-1. [Download a copy](https://github.com/openfisca/country-template/archive/master.zip) of this repository, unzip it and `cd` into it in a Terminal window.
+### Using GitHub (recommended for GitHub users)
+
+1. Click on the [“Use this template” dropdown and select “Create a new repository”](https://github.com/new?template_name=country-template&template_owner=openfisca).
+
+2. Set the repository name to `openfisca-<your_country_name>`; use underscore `_` or dash `-` as separators. For example, `openfisca-new_zealand`, `openfisca-new-zealand` or `openfisca-france`.
 
-2. Create a new repository on your favourite git host (Github, Bitbucket, GitLab, etc) with the name openfisca-your_country_name. For example, `openfisca-new_zealand` or `openfisca-france`.
+3. After being redirected to your newly created repository, please allow a short period for the automatic setup to be executed. Once done, the title of the readme should be updated to `OpenFisca <your_country_name>`.
```
MattiSG, line 19: > Can we suggest looking in the “Actions” tab to track progress?
```suggestion
3. After being redirected to your newly created repository, please allow a few minutes for the automatic setup to be executed. Once done, the title of the readme should be updated to `OpenFisca <your_country_name>`. 
```

README.md
```diff
@@ -1,25 +1,42 @@
...
+## Setting up your Country Package
 
 This set of instructions **only needs to be followed once** and will create your own copy of this boilerplate directory, customising it to the country you want to work on. You will need to have [Git](https://git-scm.com) installed.
 
-1. [Download a copy](https://github.com/openfisca/country-template/archive/master.zip) of this repository, unzip it and `cd` into it in a Terminal window.
+### Using GitHub (recommended for GitHub users)
+
+1. Click on the [“Use this template” dropdown and select “Create a new repository”](https://github.com/new?template_name=country-template&template_owner=openfisca).
+
+2. Set the repository name to `openfisca-<your_country_name>`; use underscore `_` or dash `-` as separators. For example, `openfisca-new_zealand`, `openfisca-new-zealand` or `openfisca-france`.
```
MattiSG, line 17:
```suggestion
2. Set the repository name to `openfisca-<your_country_name>`; use underscore `_` as separator if there are spaces in the country name. For example, `openfisca-new_zealand` or `openfisca-france`.
```

first-time-setup.sh
```diff
@@ -12,7 +12,7 @@ if [[ $JURISDICTION_NAME ]] && [[ $REPOSITORY_URL ]]
 then continue=Y
 fi
 
-if [[ -d .git ]]
+if [[ $CI ]] && [ ! -d .git ]
```
MattiSG, line 15: I don't understand the condition change: we now refuse to proceed if we're in CI and there is no Git history, which should never happen… but we also lose the refusal to proceed safety if we're not in CI and there is a Git history. Shouldn't this be “only if not in CI and there is a Git history”? 🤔 
```suggestion
if [[ ! $CI ]] && [[ -d .git ]]
```

MattiSG, line 15: > Is there any reason for switching from in-shell evaluation (`[[]]`) to calling the `test` program (`[]`)? 🙂

first-time-setup.sh
```diff
@@ -100,8 +103,17 @@ find . -name "*.template" -type f -delete
...
-git rm bootstrap.sh > /dev/null 2>&1
+echo -e "${PURPLE}*  ${PURPLE}Remove single use \033[0m${BLUE}first-time-setup.sh\033[0m${PURPLE} script\033[0m"
+git rm first-time-setup.sh > /dev/null 2>&1
+
+echo -e "${PURPLE}*  ${PURPLE}Remove single use \033[0m${BLUE}first-time-setup.yml\033[0m${PURPLE} GitHub Action\033[0m"
+rm -f .github/workflows/first-time-setup.yml
+
+if [[ $CI ]]
+then
+    exit 0
```
MattiSG, line 114:
```suggestion
then exit 0  # committing and tagging take directly place in the GitHub Actions workflow
```

first-time-setup.sh
```diff
@@ -62,24 +62,27 @@ last_changelog_number=$(grep --line-number '^# Example Entry' CHANGELOG.md | cut
 first_commit_message='Initial import from OpenFisca country-template'
 second_commit_message='Customise country-template through script'
 
-echo
-cd ..
-mv $parent_folder openfisca-$NO_SPACES_JURISDICTION_LABEL
-cd openfisca-$NO_SPACES_JURISDICTION_LABEL
+if [[ ! "$CI" ]]
```
MattiSG, line 65: > Unify syntax.
```suggestion
if [[ ! $CI ]]
```

.github/workflows/first-time-setup.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Set environment variables
+        run: |
+          echo "TITLE_CASE_JURISDICTION_NAME=$(echo ${{ github.event.repository.name }} | sed 's/openfisca-//' | sed 's/[-|_]/ /g' | awk '{for (i=1; i<=NF; i++) $i=toupper(substr($i,1,1)) substr($i,2)} 1')" >> $GITHUB_ENV
+
+      - name: Execute the first-time-setup script
+        run: CI=true JURISDICTION_NAME="$TITLE_CASE_JURISDICTION_NAME" REPOSITORY_URL="${{ github.repositoryUrl }}" ./first-time-setup.sh
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
```
MattiSG, line 31: Latest version is v5, any reason to stick to v4? 🙂

.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
+name: Build
+
+on:
+  workflow_call:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
Ndpnt: Since the previous monolithic workflow had this option, I included it here. However, you're right—it doesn't seem to be very useful, particularly because this workflow is a dependency of others and there's no reason to call it on its own.

first-time-setup.sh
Ndpnt: My mistake.

.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9.12 # Patch version must be specified to avoid any cache confusion, since the cache key depends on the full Python version. Any difference in patches between jobs will lead to a cache not found error.
+      - name: Restore build
```
MattiSG, line 16: This is the one step where we actually create the build 🙂 
```suggestion
      - name: Cache build
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
+name: Deploy
+
+on:
+  pull_request_target:
+    branches: 
+      - main
+    types: [ closed ]
```
MattiSG, line 7: > Unify syntax.
```suggestion
    branches: [ main ]
    types: [ closed ]
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  validate:
+    uses: "./.github/workflows/validate.yml"
+
+  # GitHub Actions does not have a halt job option to stop from deploying if no functional changes were found.
+  # We build a separate job to substitute the halt option.
+  # The `deploy` job is dependent on the output of the `check-for-functional-changes` job.
```
MattiSG, line 15:
```suggestion
  # We thus execute a separate deployment job depending on the output of this job.
```

.github/workflows/deploy.yml
MattiSG, line 7: Shouldn't we deploy on any push to `main`, even if they did not originate from a merged pull request? 🤔 
This would ensure a consistent mapping between commits on `main` and deployments, and simplify the code by removing the need to check `if: github.event.pull_request.merged == true`.

If changed, this should be reflected in the CHANGELOG.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9.12
+      - id: stop-early
+        run: if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh" ; then echo "::set-output name=status::success" ; fi
+
+  deploy:
+    runs-on: ubuntu-22.04
+    needs: [ validate, check-for-functional-changes ]
+    if: github.event.pull_request.merged == true && needs.check-for-functional-changes.outputs.status == 'success'
```
MattiSG, line 35: > This condition is redundant since `check-for-functional-changes` already checks that `github.event.pull_request.merged == true`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      PYPI_TOKEN_OPENFISCA_BOT: ${{ secrets.PYPI_TOKEN_OPENFISCA_BOT }}
```
MattiSG, line 37: Do we really need to set this variable in `env`, can't we directly interpolate `${{ secrets.PYPI_TOKEN_OPENFISCA_BOT }}` in the upload step? 🙂

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9.12
+      - name: Restore build
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.pythonLocation }}
+          key: build-${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}-${{ github.sha }}
+      - name: Cache release
```
MattiSG, line 49:
```suggestion
      - name: Restore built package
```

MattiSG: > Though outside of scope of this PR, this step should definitely fail on cache miss!

.github/workflows/first-time-setup.yml
```diff
@@ -0,0 +1,36 @@
...
+  first-time-setup:
+    # Ensure this job does not run on the template repository or when the repository is forked
+    if: ${{ !github.event.repository.is_template && !github.event.repository.fork }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
```
MattiSG, line 22: > Use consistent line breaks across workflows 🙂 Others currently have no empty line between `steps`, either remove the ones here or add them in the others.

.github/workflows/first-time-setup.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Set environment variables
```
MattiSG, line 23:
```suggestion
      - name: Infer jurisdiction name from repository name
```

README.md
```diff
@@ -1,25 +1,42 @@
...
 
-1. [Download a copy](https://github.com/openfisca/country-template/archive/master.zip) of this repository, unzip it and `cd` into it in a Terminal window.
+### Using GitHub (recommended for GitHub users)
+
+1. Click on the [“Use this template” dropdown and select “Create a new repository”](https://github.com/new?template_name=country-template&template_owner=openfisca).
+
+2. Set the repository name to `openfisca-<your_country_name>`; use underscore `_` as separator if there are spaces in the country name. For example, `openfisca-new_zealand` or `openfisca-france`.
 
-2. Create a new repository on your favourite git host (Github, Bitbucket, GitLab, etc) with the name openfisca-your_country_name. For example, `openfisca-new_zealand` or `openfisca-france`.
+3. After being redirected to your newly created repository, please allow a few minutes for the automatic setup to be executed. Once done, the title of the readme should be updated to `OpenFisca <your_country_name>`.
```
MattiSG, line 19:
```suggestion
3. After being redirected to your newly created repository, please allow a few minutes for the automatic setup to be executed. Once done, the title of the README file should be updated to `OpenFisca <your_country_name>`. 
```

first-time-setup.sh
```diff
@@ -100,8 +103,16 @@ find . -name "*.template" -type f -delete
 echo -e "${PURPLE}*  ${PURPLE}Rename package to: \033[0m${BLUE}$package_name\033[0m"
 git mv openfisca_country_template $package_name
 
-echo -e "${PURPLE}*  ${PURPLE}Remove single use \033[0m${BLUE}bootstrap.sh\033[0m${PURPLE} script\033[0m"
-git rm bootstrap.sh > /dev/null 2>&1
+echo -e "${PURPLE}*  ${PURPLE}Remove single use \033[0m${BLUE}first-time-setup.yml\033[0m${PURPLE} GitHub Action\033[0m"
+git rm .github/workflows/first-time-setup.yml > /dev/null 2>&1
+
+echo -e "${PURPLE}*  ${PURPLE}Remove single use \033[0m${BLUE}first-time-setup.sh\033[0m${PURPLE} script\033[0m"
```
MattiSG, line 109: > Let's group them and reduce the amount of logs we expose our users to 🙂 
```suggestion
echo -e "${PURPLE}*  ${PURPLE}Remove single use first time setup files\033[0m"
git rm .github/workflows/first-time-setup.yml > /dev/null 2>&1
```

MattiSG (commented):
Beautiful! 🤩 
Thank you for this cloud first time setup, and for the overall upgrades to the CI pipeline! I could have stayed more focused on the main goal of this changeset in my comments, apologies for that.
It is the same with the bunch of comments I make here now. I will apply some of the changes in wording myself, especially when they touch upon pre-existing content.

One question though: I see that the current workflow layout has a nested `build` job, and it is the only nested one. Do you understand where this is coming from? Is there any way we can make it appear at the same level as other entries in the build?

image

MattiSG:
I have switched off required status for all status checks except `check-version-and-changelog` 👍

Ndpnt:
> One question though: I see that the current workflow layout has a nested `build` job, and it is the only nested one. Do you understand where this is coming from? Is there any way we can make it appear at the same level as other entries in the build?
> 
> image

I'm unsure why, and I don't know how to change it. My guess is that this arises because the `build` job is defined within the `build` workflow file, and this job is triggered only as a dependency of other workflows, not directly.

MattiSG:
> My guess is that this arises because the `build` job is defined within the `build` workflow file, and this job is triggered only as a dependency of other workflows, not directly.

That makes sense… let's try to find a workaround this UI limitation and rename one or the other to make it look clearer 🙂

Ndpnt:
> That makes sense… let's try to find a workaround this UI limitation and rename one or the other to make it look clearer 🙂

What do you think if we rename the job `build` into `build-and-cache`?

MattiSG:
> rename the job `build` into `build-and-cache`?

Very good idea, as indeed the goal is to cache and fill in the dependency for all other steps!
Let's make sure that this is the name of the top-level entry though, so I assume this should be the name of the _workflow_ and not _job_ (but I'm not sure).

Ndpnt:
> Very good idea, as indeed the goal is to cache and fill in the dependency for all other steps! Let's make sure that this is the name of the top-level entry though, so I assume this should be the name of the _workflow_ and not _job_ (but I'm not sure).

I'm not sure to understand, what do you prefer `build / build-and-cache` or `build-and-cache / build`?

MattiSG:
Thanks for pointing that out… I prefer `build / build-and-cache` even though I initially suggested the opposite 😅

MattiSG (commented):
The code LGTM, but I spot something odd in the workflow view when applying this changeset, where the dependencies seem to be wrong.

### [Before](https://github.com/openfisca/country-template/actions/runs/8920375488)

image


### [After](https://github.com/openfisca/country-template/actions/runs/8941054687)

image

MattiSG:
As discussed synchronously:

- The parallelisation between `check-changelog` and `build` is voluntary and I had already suggested it, in order to shorten the feedback loop for users.
- The `deploy` tasks are now only triggered on `main` and thus voluntarily do not appear on other branches. The dependency layout on a deployment branch can be seen on the [source repository](https://github.com/Ndpnt/country-template/actions/runs/8941054081) and below.

image

MattiSG (approved):
🚀

PR #147, opened 2024-05-09
Skip `deploy` workflow when PyPI token is not defined in GitHub secrets

Ndpnt:
* Technical improvement
* Impacted areas: CI workflows
* Details:
  - Skip `deploy` workflow when PyPI token is not defined in GitHub secrets
  - Rename the GitHub secret `PYPI_TOKEN_OPENFISCA_BOT` used in `deploy` workflow to `PYPI_TOKEN`
  - Ensure lint scripts work properly without reachable tags
  - Update deprecated syntax in GitHub Action workflow


.github/workflows/deploy.yml
```diff
@@ -23,12 +23,29 @@ jobs:
         with:
           python-version: 3.9.12
       - id: stop-early
-        run: if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh" ; then echo "::set-output name=status::success" ; fi
+        run: |
+          if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh" ; then 
+            echo "status=success" >> $GITHUB_OUTPUT
```
MattiSG, line 28: If we go multiline, let's do it all the way 😉 
```suggestion
          if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh"
          then echo "status=success" >> $GITHUB_OUTPUT
```

.github/workflows/deploy.yml
```diff
@@ -23,12 +23,29 @@ jobs:
...
 
+  check-pypi-token:
+    runs-on: ubuntu-22.04
+    outputs:
+      pypi_token_present: ${{ steps.check_token.outputs.pypi_token_present }}
+    steps:
+      - name: Check PYPI Token
+        id: check_token
+        run: |
+          if [[ -n "${{ secrets.PYPI_TOKEN }}" ]]; then
```
MattiSG, line 39: > The comment you made in [the commit](https://github.com/openfisca/country-template/pull/147/commits/7afe3008235bfcbf04d30fdc9c5567eb664059a7) to explain the implementation  is really good, but it would be useful to also have it as a comment in the code to save some future smartass 😉

.github/workflows/deploy.yml
```diff
@@ -23,12 +23,29 @@ jobs:
...
+          if "${GITHUB_WORKSPACE}/.github/has-functional-changes.sh" ; then 
+            echo "status=success" >> $GITHUB_OUTPUT
+          fi
 
+  check-pypi-token:
+    runs-on: ubuntu-22.04
+    outputs:
+      pypi_token_present: ${{ steps.check_token.outputs.pypi_token_present }}
+    steps:
+      - name: Check PYPI Token
```
MattiSG, line 36:
```suggestion
      - name: Check PyPI token is defined
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+### 7.1.2 [#147](https://github.com/openfisca/country-template/pull/147)
+
+* Technical improvement
+* Details:
+  - Skip `deploy` workflow when PyPI token is not defined in GitHub secrets
+  - Ensure lint scripts work properly without reachable tags
+  - Rename the GitHub secret `PYPI_TOKEN_OPENFISCA_BOT` used in `deploy` workflow to `PYPI_TOKEN`
+  - Update deprecated syntax in GitHub Action workflow
```
MattiSG, line 10:
```suggestion
  - Update deprecated syntax in GitHub Actions workflow
```

README.md
```diff
@@ -56,7 +56,35 @@ The files that are outside from the `openfisca_country_template` folder are used
 
 ## Packaging your Country Package for Distribution
 
-Country packages are python distributions. To distribute your package via `pip`, follow the steps given by the [Python Packaging Authority](https://python-packaging-user-guide.readthedocs.io/tutorials/distributing-packages/#packaging-your-project).
+Country packages are Python distributions. You can choose to distribute your package automatically via our continuous deployment system on GitHub, or manually.
```
MattiSG, line 59:
```suggestion
Country packages are Python distributions. You can choose to distribute your package automatically via the predefined continuous deployment system on GitHub Actions, or manually.
```

README.md
```diff
@@ -56,7 +56,35 @@ The files that are outside from the `openfisca_country_template` folder are used
...
+
+This involves detailed steps on preparing your package, creating distribution files, and uploading them to PyPI.
+
+### Accessing your package
+
+Once deployed, your package can be easily installed by users with the following command:
+
+```sh
+pip install openfisca-<your_country_name>
+```
```
MattiSG, line 87: This is redundant with the instructions [below](https://github.com/openfisca/country-template/pull/147/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R126-R142) 🙂 
```suggestion
```

MattiSG (changes requested):
Excellent work!
However, in my testing, `git describe --always` falls back to `HEAD` in case there are no tags, meaning that in such a case we will never lint any file. I understand this is better than failing, but I believe we can come up with a better alternative in case there are no tags, such as using the GitHub-provided PR target, or maybe `git rev-list --max-parents=0 HEAD` to get the root commit and thus lint _all_ files.

MattiSG:
We will need @sandcha to create a new PyPI token too, to set the new `PYPI_TOKEN` variable 😅

MattiSG (approved):
Beautiful 🤩 
Let's merge without `PYPI_TOKEN` defined to confirm that deployment is cleanly skipped, and then define it and re-run the job to confirm that deployment is activated 🙂

PR #148, opened 2024-05-12
Update pyproject.toml to include licence and reorder to match officia…

verbman:
* Technical improvement.
* Minor
* Details:
  - Add licence to pyproject.toml, reorder to match official pyproject.toml example

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,16 @@
 [project]
 name = "openfisca-country_template"
-version = "7.1.2"
-description = "OpenFisca Rules as Code model for Country-Template."
-readme = "README.md"
-keywords = ["microsimulation", "tax", "benefit", "rac", "rules-as-code"]
+version = "7.1.3"
+dependencies = [
+    "openfisca-core[web-api] >=41.4.5, <42.0.0"
+]
+requires-python = ">=3.9"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,11 +1,16 @@
...
+version = "7.1.3"
+dependencies = [
+    "openfisca-core[web-api] >=41.4.5, <42.0.0"
+]
+requires-python = ">=3.9"
+authors = []
+maintainers = []
 description = "OpenFisca Rules as Code model for Country-Template."
 readme = "README.md"
+license = {file = "LICENSE"}
```
MattiSG, line 12: > If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with License ::.

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

MattiSG (commented):
Not sure this is a good idea…

MattiSG:
To clarify: the licence information is redundant and according to the doc we should use classifiers. If you want to get the PR to focus on reordering `pyproj` to follow recommendations I'm happy to merge that:)

PR #152, opened 2024-05-19
Update classifiers for supported versions of python in pyproject.toml

verbman:
* Technical improvement. 
* Details:
  - Update classifiers in pyproject.toml to include all supported versions of python


MattiSG (approved):
LGTM! Could be relevant to integrate the reordering of metadata that didn't make it in #148 🙂

PR #158, opened 2024-10-16
build: test built version

bonjourmauko:
Part of https://github.com/openfisca/openfisca-core/issues/1064
Depends on #156 

* Technical Changes
* Details:
  - Run tests in absolute isolation
    - Uses `tox` to test builds in isolation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ uninstall:
 clean:
 	rm -rf build dist
 	find . -name '*.pyc' -exec rm \{\} \;
+	find . -type d -name '__pycache__' -exec rm -r {} +
 
 deps:
 	pip install --upgrade pip build twine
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -24,30 +25,24 @@ build: clean deps
...
-check-syntax-errors:
-	python -m compileall -q .
-
-format-style:
+format:
 	@# Do not analyse .gitignored files.
 	@# `make` needs `$$` to output `$`. Ref: http://stackoverflow.com/questions/2382764.
-	ruff format `git ls-files | grep "\.py$$"`
-	isort `git ls-files | grep "\.py$$"`
 	black `git ls-files | grep "\.py$$"`
```
benoit-cty, line 31: Why keeping Black and isort when Ruff says "Drop-in parity with Flake8, isort, and Black" ?

benoit-cty (approved):
Thanks for all the work.

Maybe it will be nice to add some installation instructions in the contributing.md :

You need to have tox and Python 3.9 to run test like the CI does :
```
# Install PyEnv
curl https://pyenv.run | bash
# Install Python 3.9 needed by tox
pyenv install 3.9.20
# Install pipx
sudo apt install pipx
# Install tox
pipx install tox
```
Run lint with `tox -r -e lint`
Run test with `tox -r -e py39`

Makefile
bonjourmauko: For isort, because of heavy customisation. But black can go, en effect.

PR #163, opened 2025-01-06
Fix setup script

### LINKED ISSUE #162: .flake8 presence breaks first-time-setup.sh script
When the .flake8 file is not present (as is generally the case on new builds), the ./first-time-setup.sh script fails. Removing .flake8 solves the issue 
https://github.com/openfisca/country-template/blob/5c33a5c0930d3ee83544b9f4d6a7d84b2f39d8c3/first-time-setup.sh#L84

jganseman:
* Crash fix.
* Impacted areas: `./first-time-setup.sh`
* Details:
  - Remove mention of .flake8 which crashes setup script when not present

- - - -

These changes :

- Fix or improve an already existing calculation --> setup script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 7.1.9 [#163](https://github.com/openfisca/country-template/pull/163)
+
+* Crash fix
+* Impacted areas: `first-time-setup.sh`
+* Details:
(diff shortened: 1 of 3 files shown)


MattiSG (approved):
Thanks!

MattiSG:
We just need to bump the version number (patch bump) and update the CHANGELOG.

jganseman (commented):
changelog and version updated

PR #165, opened 2025-03-03
feat: correctly implement disposable income

### LINKED ISSUE #164: Fix formula for disposable income
### Context

Disposable income is supposed to include all incomes and taxes and benefits.

### Problem

It only includes salary and taxes and basic income, excluding capital returns, pensions, and social benefits.

bonjourmauko:
_Note: changes break tests in core. Please do not merge immediatly._

Fixes #164

* Impacted periods: from 01/01/2015.
* Impacted areas: `disposable_income`
* Details:
  - Calculate `disposable_income` for the entire household
  - Include taxes and benefits in `disposable income`
  - Allow for triplets in households

- - - -

These changes:

- Add a new feature (for instance adding a variable)
- Fix or improve an already existing calculation.

### DIFF
--- a/openfisca_country_template/entities.py
+++ b/openfisca_country_template/entities.py
@@ -14,7 +14,7 @@ Household = build_entity(
     label="All the people in a family or group who live together in the same place.",
     doc="""
     Household is an example of a group entity.
-    A group entity contains one or more individual·s.
+    A group entity contains one or more individuals.
     Each individual in a group entity has a role (e.g. parent or children).
     Some roles can only be held by a limited number of individuals (e.g. a
     'first_parent' can only be held by one individual), while others can
@@ -26,7 +26,7 @@ Household = build_entity(
 
     Usage:
         Check the number of individuals of a specific role (e.g. check if there
-        is a 'second_parent' with household.nb_persons(Household.SECOND_PARENT)).
+        is a 'main' adult with household.nb_persons(Household.MAIN)).
         Calculate a variable applied to each individual of the group entity
         (e.g. calculate the 'salary' of each member of the 'Household' with:
             salaries = household.members("salary", period = MONTH)
@@ -36,24 +36,22 @@ Household = build_entity(
     """,
     roles=[
         {
-            "key": "cohabiting_partner",
-            "plural": "cohabiting_partners",
-            "label": "Cohabiting Partners",
-            "subroles": ["main_partner", "other_partner", "another_partner"],
-            "doc": "The one or many adults in charge of the household.",
+            "key": "main",
+            "label": "Main",
+            "max": 1,
+            "doc": "The main person of the household for calculation purposes.",
         },
         {
-            "key": "parent",
-            "plural": "parents",
-            "label": "Parents",
-            "subroles": ["first_parent", "second_parent", "other_parent"],
-            "doc": "The one or many adults in charge of the household.",
+            "key": "adult",
+            "plural": "adults",
+            "label": "Adult",
+            "doc": "The other adults of the household.",
         },
         {
             "key": "child",
             "plural": "children",
             "label": "Child",
-            "doc": "Other individuals living in the household.",
+            "doc": "The other non-adult individuals living in the household.",
         },
     ],
 )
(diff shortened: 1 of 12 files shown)


openfisca_country_template/entities.py
```diff
@@ -35,13 +35,19 @@
...
+        },
         {
             "key": "parent",
             "plural": "parents",
             "label": "Parents",
-            "max": 2,
-            "subroles": ["first_parent", "second_parent"],
-            "doc": "The one or two adults in charge of the household.",
+            "subroles": ["first_parent", "second_parent", "other_parent"],
+            "doc": "The one or many adults in charge of the household.",
```
sylvainipp, line 50: I agree with the idea, but it seems to me it introduces a bit of uncertainty : who is supposed to be a partner and who a parent ? Here, the uncertainty comes from the fact that we don't have a legislation to guide us. In France for instance, we have a family defined for some benefits, that doesn't allow for more than two parents, and households that allow for as many adults as we want. Here, separating the roles ('partner' and 'parent') but making them exclusive is maybe confusing. We can either create a new entity, one for family/tax household with a constraint on the number of adults and one for the households with no constraint, or decide clearly this is a household, and thus only keeping one adult main role with an unlimited number of persons after the first subrole. On the name of this role, both `partner`, `parent `and `adult  ` are good from my point of view (in the French and Senegalese models at least, we call 'parent' adults without children, and it doesn't look like a big issue to me). Moreover, if this change break things on core tests, maybe creating a new entity is too much work if you want to present it in two weeks.

sylvainipp (commented):
I am a bit sceptical on the difference cohabiting_partner/parent for the same entity, but otherwise everything is good for me

openfisca_country_template/entities.py
sarafalamaki: What if there's a household with one parent and they have a cohabiting partner who is not the children's parent?

bonjourmauko:
I'm not sure also how to describe spouses with or without kids. Spouses?

sylvainipp:
In my opinion, we need to decide which entity we are speaking of and to stick to this. If this is a household, we don't need the marital status of the adults, but only a way of having one of them that is identified (to help for some computations), for instance `adult1`, the other adults with a subrole like `other_adult`, and the child (the difference between adults and child is useful to compute standard of living for instance). If no other characteristic is used in the tax and benefit system, we don't have to specify them.

bonjourmauko:
I dedided to avoid the parenting issue completely here, as the scope of the changes is with the calculation of `disposable_income`. If in the future we want to show how these can differ, as per @sarafalamaki 's example, say on `parenting_allowance`, we can easily complete the code example with a new group entity for that.

openfisca_country_template/entities.py
```diff
@@ -36,18 +36,22 @@
     """,
     roles=[
         {
-            "key": "parent",
-            "plural": "parents",
-            "label": "Parents",
-            "max": 2,
-            "subroles": ["first_parent", "second_parent"],
-            "doc": "The one or two adults in charge of the household.",
```
sylvainipp, line 44: I think it will be good to have an exemple of subroles in the country-template, as it may be useful in many tax benefit systems, maybe having a role adule with subroles 'main' and 'other_adult' could be good ?
And sorry for all the remarks coming one by one, I have no clear general ideas of what we want at the end.

bonjourmauko: The problem is, with the current core implementation, it limits the number of role members. So for example, if there are first and second parent, there can't be no third. That is whyt I did it the way I did in the previous iteration.

openfisca_country_template/variables/benefits.py
```diff
@@ -125,7 +125,13 @@ def formula(household, period, parameters):
...
         income_condition = household_income <= income_threshold
 
-        is_single = household.nb_persons(Household.PARENT) == 1
+        is_single = (
+            household.nb_persons(Household.MAIN) == 1
+            and household.nb_persons(Household.ADULT) == 0
+        ) or (
+            household.nb_persons(Household.MAIN) == 0
+            and household.nb_persons(Household.ADULT) == 1
+        )
```
sylvainipp, line 134: Don't we want to always have a main ?

sylvainipp (approved):
I'm still not comfortable to have this possibility of only an adult or only a main in the household, knowing that in practice there are no real difference. But if you want to merge, I have no complete opposition or solution easy to implement.

bonjourmauko:
> I'm still not comfortable to have this possibility of only an adult or only a main in the household, knowing that in practice there are no real difference. But if you want to merge, I have no complete opposition or solution easy to implement.

I can remove the main. There is no actual need for it in the demo.

sylvainipp:
I think it will be more suited to help for the creation of new tax and benefit systems, but I let you decide

bonjourmauko:
> I think it will be more suited to help for the creation of new tax and benefit systems, but I let you decide

This is a living base for new systems, we can always add new examples later :)

PR #169, opened 2026-01-02
Add an example of an inline parametric reform

guillett:
Related to https://github.com/openfisca/openfisca-core/pull/1334


benjello (approved):

guillett:
I realized that there are issues with `uv`, `uv.lock` and the CI.

I'm getting my head around it with another package.

We may prefer to merge another cleaning PR before merging functional changes.

